PR #3, opened 2017-02-15
Removed k8s deployments limits and user quota

xoen:
We don't know yet how resources are manages and how much resources our
users will use.

PS: I'll also remove the LIMITS_CPU/LIMITS_MEMORY params in the Jenkins
    config.


xoen:
@kerin FYI - This should be enough, let me know if I'm missing something.

PR #8, opened 2017-02-21
Users can list buckets in AWS Console (S3 policies)

xoen:
From the AWS documentation:

> To list all buckets, users require the GetBucketLocation and ListAllMyBuckets actions for all resources in Amazon S3

Unfortunately **this will show the users all the buckets** (but they'll only
have access to the source/scratch one). From some of the comments on the
internet show only some of the buckets was not supported (yet).
But there may be a solution for this.

See: https://aws.amazon.com/blogs/security/writing-iam-policies-how-to-grant-access-to-an-amazon-s3-bucket/

Ticket: https://trello.com/c/yKYfs7cf

### DIFF
--- a/infra/terraform/modules/data_buckets/iam.tf
+++ b/infra/terraform/modules/data_buckets/iam.tf
@@ -1,3 +1,10 @@
+# - The s3:GetBucketLocation and s3:ListAllMyBuckets actions are necessary
+#   to allow the user to see the buckets in the AWS console
+# - The s3:ListBucket action is necessary to allow users to list objects
+#   in a bucket
+#
+# See: https://aws.amazon.com/blogs/security/writing-iam-policies-how-to-grant-access-to-an-amazon-s3-bucket/
+
 resource "aws_iam_group" "managers" {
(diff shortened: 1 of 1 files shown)


xoen:
Hey @kerin this should allow the users to list **all** the buckets. I'll do more reading about S3 policies to see if there is some way to show users only some buckets but it's not clear if this is even possible at the moment.

xoen:
@kerin I also simplified the IAM policies by merging some of the statements and make the files simpler.

I didn't find a way (not sure there is one!) to only show buckets they can access so after lunch I'll merge this and apply unless you find something wrong in this and/or you have any feedback for me.

PS: I want to read this ["IAM: Access to User-Specific Folders in S3"](https://aws.amazon.com/blogs/security/writing-iam-policies-grant-access-to-user-specific-folders-in-an-amazon-s3-bucket/) article but after that I'll declare the case close for now.

xoen:
FYI: Applied and tested with my *AldoAnalyst* user and it seems to work, so merging it.

PR #9, opened 2017-02-23
Enforce S3 encryption on upload

xoen:
See: http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .terraform
 *.tfstate*
 *.tgz
+.DS_STORE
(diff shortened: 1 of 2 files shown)


xoen:
I added these policies to the buckets rather than to the groups because I realised it's safer: e.g. if we'll add another user group and we forget to add these rules it would allow users to upload unencrypted files.

xoen:
Hey @kerin could you have a look at this please (***not urgent***) and let me know if I can merge/terraform apply it?

PR #10, opened 2017-02-27
Added AWS lambda to encrypt S3 objects after upload

xoen:
4 terraform resources are needed:
- The lambda function itself
- The role running the lambda function
- The notification from the S3 bucket when an S3 object is created
- The permission for the bucket to invoke the lambda function

I automated the lambda function packaging by using terraform's [`data "archive_file"`](https://www.terraform.io/docs/providers/archive/d/archive_file.html). The `aws_lambda_function` resource depends on it. This resource has also a `source_code_hash` attribute with its value depending on the zip archive file (to force a re-deploy when the lambda function change).


xoen:
FYI @kerin These are the changes I made (I thought you may be interested to see them before a merge/rebase)

PR #17, opened 2017-03-06
RStudio: Converted existing k8s templates to Helm Chart

xoen:
NOTE: Need to fill values for alpha/dev environment with credentials/config


charts/rstudio/templates/deployment.yml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: r-studio
```
kerin: Resource names should be populated from the 'fullname' template helper, e.g. `name: {{ template "fullname" . }}`

charts/rstudio/templates/deployment.yml
```diff
@@ -4,6 +4,7 @@ metadata:
...
   labels:
     app: r-studio
```
kerin: `app` should also use the template fullname, eg `app: {{ template "fullname" . }}`

charts/rstudio/templates/ingress.yml
```diff
@@ -2,11 +2,14 @@ apiVersion: extensions/v1beta1
...
   name: r-studio
+  labels:
+    app: r-studio
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
   annotations:
     kubernetes.io/ingress.class: "nginx"
 spec:
   rules:
-  - host: {{.Username}}-rstudio.{{.ToolsDomain}}
+  - host: {{ .Values.Username }}-rstudio.{{ .Values.ToolsDomain }}
```
xoen: Also, I think this is actually `$USER.rstudio.$TOOLS_DOMAIN` (dot instead of dash).

But how did this work before?

charts/rstudio/README.md
```diff
@@ -0,0 +1,33 @@
+# RStudio Helm Chart
```
xoen: I added a basic README, so we can ideally almost Copy & Paste from it when installing it for a user. Ideally we have a `README.md` in each chart directory.

PR #20, opened 2017-03-09
k8s Dashboard authentication/authorisation

xoen:
The k8s Dashboard didn't require any authorisation. I added an auth-proxy container in front of it.

The auth-proxy uses Auth0 for authentication. It also checks that the user's nickname is matching the given regular expression (`NICKNAME_RE` environment variable).

See [analytics-platform-auth-proxy](https://github.com/ministryofjustice/analytics-platform-auth-proxy/blob/master/routes/index.js) container image for more details.


xoen:
Have a look if you want @kerin :)

It should work now: http://dashboard.services.alpha.mojanalytics.xyz

kerin:
LGTM

PR #21, opened 2017-03-14
Helm Charts: Documented steps and updated config for 'dev' environment

xoen:


xoen:
@kerin see if the order and the commands are OK and feel free to tweak the readme if you want.

xoen:
(GH diff is not very useful in this case but it's mainly about adding clients to OAuth, changing some of the URLs or things like that)

PR #22, opened 2017-03-16
Storing username/email and fullname in user secrets

xoen:
This secret is in the user's namespace.


xoen:
Hey @kerin see if there is anything wrong here, especially with the bash quoting. I tested the secret parts locally, and I'm going to merge/test this on Jenkins as well to see if it works end-to-end.

PR #23, opened 2017-03-16
config-user helm chart

xoen:
This will set up things like git config and other user-specific
things (in the future).

For the moment this will just create the ~/.gitconfig file.


xoen:
Merging this as discussed offline.

PR #27, opened 2017-03-29
Tweaked shiny-app helm chart to not use git-sync

xoen:
Instead of mounting the shiny app code as a container we'll use a Docker
image which already has the shiny app (build with Shippable or whatever).


charts/shiny-app/values.yaml
```diff
@@ -6,15 +6,6 @@ app:
 shinyApp:
   port: 3838
   docker:
-    repository: rocker/shiny
+    repository: ""
     tag: latest
```
kerin: We should have this templated too, as Shippable should always be generating a build-specific image tag

xoen: You're right, it's always better to specify the tag rather than use `latest` which can be unpredictable. As discussed it's possible to override the tag, I documented the value in the README but I omitted it in the run command for simplicity sake. But I'll update it now.

xoen: Updated: https://github.com/ministryofjustice/analytics-platform-ops/pull/27/commits/c5220bbfbf4577603e3f0861fef167efca3899e6

xoen:
Merging this @kerin

PR #30, opened 2017-04-06
IAM Policies to restrict access to CREST S3 bucket

xoen:
The `${ENV}-moj-analytics-crest` S3 bucket contains sensitive information.

Users in the `${ENV}-crest-manager` can read/write in it, everyone else can't access it.

**NOTE**: Note that this is currently work in progress.


xoen:
Almost forgot...need to add the lambda function to encrypt the crest bucket S3 objects also to the `alpha` environment.

PR #45, opened 2017-05-04
Publish GitHub webhook events to SNS

xoen:
### What it does
Provides a lambda function which publishes [GitHub webhook](https://developer.github.com/v3/repos/hooks/#create-a-hook) events to SNS (to the relevant topic).

### Details
We want environments to be separated so resources are namespaced by starting with the environment name, e.g. `dev_foo` or `alpha_bar`. This is also true of the SNS topics: different environments will have separate topics.

This function will be triggered by an AWS API Gateway endpoint (`POST /github`) with _Lambda Proxy integration_ enabled because we need to have access to the request headers.

GH will provide the event type in the `X-GitHub-Event` header, e.g. `team`, `membership`, etc...
The body of the web hook notification/request (found in `event["body"]`) will be published to the SNS topic `$ENV_gh_$TYPE_events` (e.g. `alpha_gh_team_events`)

### Authentication
The endpoint triggering this lambda function doesn't require authentication. 
The lambda function will have to check the validity of the signature in the request (`X-Hub-Signature` header, containing the HMAC hex digest of the body).

### Big picture
This is part of providing access to sensitive data in S3 buckets.
We want to listen to GH events related to teams/members to update AWS IAM roles/policies accordingly.

### Ticket
[Create webhook handler Lambda function to receive from API Gateway and publish to SNS queues](https://trello.com/c/UG1QR9gX/114-create-webhook-handler-lambda-function-and-connect-to-api-gateway)

### TODO
- [x] Verify GitHub signature
- [x] Tidy up
- [x] Use `data_access` module in _alpha_ environment
- [x] Add some tests


xoen:
test comment

infra/terraform/modules/lambda_functions/encryption.tf
```diff
@@ -9,13 +9,13 @@ data "archive_file" "lambda_function_package" {
 resource "aws_lambda_function" "encrypt_s3_object" {
     description = "Encrypt S3 objects using AWS' server side encryption"
     filename = "/tmp/encrypt_s3_object.zip"
-    source_code_hash = "${data.archive_file.lambda_function_package.output_base64sha256}"
+    source_code_hash = "${data.archive_file.encrypt_s3_object_zip.output_base64sha256}"
```
xoen: Test comment

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+        message
+    ))
+
+    try:
+        sns.publish(
+            TopicArn=topic_arn,
+            Message=message
+        )
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "NotFound":
```
xoen: Boto3 error handling is somehow low level, that's why I wrapped it in a custom `SNSTopicNotFoundError` exception, to simplify the client code a little bit.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+        topic_arn,
+        message
+    ))
+
+    try:
+        sns.publish(
+            TopicArn=topic_arn,
+            Message=message
+        )
+    except ClientError as e:
```
andyhd: Avoid using single character variable names except for iterators. This should be `error`.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+        if e.response["Error"]["Code"] == "NotFound":
+            err = "SNS topic '{}' not found".format(topic_arn)
```
andyhd: This error message duplicates information already contained in the exception class name. All that is needed is to pass the not found topic `topic_arn` to the exception constructor.
```python
raise SNSTopicNotFound(topic_arn)
```

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+    print("Response {}: '{}'".format(status_code, message))
+
+    return {
+        "statusCode": status_code,
+        "headers": {"Content-Type": "application/json"},
+        "body": json.dumps({"message": message})
+    }
+
+
+# See article on handling GH Webhooks in Python:
```
andyhd: Maybe more information on what this article has to do with the code is required. The title says it's about Django, which we're not using, so it's a bit confusing.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+# https://simpleisbetterthancomplex.com/tutorial/2016/10/31/how-to-handle-github-webhooks-using-django.html
+def _valid_signature(hmac_signature, message):
+    sha_name, signature = hmac_signature.split('=')
+    if sha_name != "sha1":
+        return False
```
andyhd: If the logic is reversed, only 2 return statements are needed in this function, eg:
```python
if sha_name == 'sha1':
    mac = hmac.new(...)
    if hmac.compare_digest(...):
        return True
return False
```

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+
+    mac = hmac.new(
+        GH_HOOK_SECRET.encode(),
```
andyhd: The default encoding is UTF-8, but it is better to be explicit. Eg: `GH_HOOK_SECRET.encode('utf-8')`

infra/terraform/environments/alpha/main.tf
```diff
@@ -82,3 +82,10 @@ module "container_registry" {
     source = "../../modules/container_registry"
     env = "${var.env}"
 }
+
+module "data_access" {
+    source = "../../modules/data_access"
+    env = "${var.env}"
+    sns_arn = "${var.sns_arn}"
+    gh_hook_secret = "${var.gh_hook_secret}"
+}
```
xoen: Test comment (to trigger the web hook)

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+        msg=message.encode(),
+        digestmod=sha1
+    )
+    if hmac.compare_digest(mac.hexdigest().encode(), signature.encode()):
+        return True
+
+    return False
+
+
+def handler(event, context):
```
andyhd: `handler` is a really generic name. Give this function a name that reveals its intent.

infra/terraform/environments/alpha/main.tf
xoen: another comment

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+    print("event received = {}".format(event))
+
+    topic = event["headers"]["X-GitHub-Event"]
+    topic_arn = _topic_arn(topic)
```
andyhd: The `topic` variable is only used once immediately after it is set, so it can be inlined.

andyhd: If the value is inlined, then the `topic` variable can be set to the result of `_topic_arn()` and the `_publish(topic, message)` line will be much more straightforward.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: Also, this constant `GH_HOOK_SECRET` is only used with this encoding, so it can be encoded when first set, eg:
```python
GH_HOOK_SECRET = os.environ['GH_HOOK_SECRET'].encode('utf-8')
```

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
+import hmac
+from hashlib import sha1
+import json
+import os
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+SNS_ARN = os.environ.get("SNS_ARN")
```
andyhd: If the `SNS_ARN` environment variable is not set, this code will not complain.
The default value for `os.environ.get("SNS_ARN")` if the named item is not found is `None`. This may cause unexpected behaviour - publishing messages to a topic named `None:dev_gh_team_events`, for example.
Use dict access (`os.environ['SNS_ARN']`) instead so that the error will be raised at this point and will be self-explanatory.

andyhd:
How are we unit testing this code?

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+STAGE = os.environ.get("STAGE")
+GH_HOOK_SECRET = os.environ.get("GH_HOOK_SECRET")
+
+sns = boto3.client("sns")
+
+
+class SNSTopicNotFoundError(Exception):
```
xoen: Funny how I didn't include the `Error` suffix in the first place and then when I looked at the [Python documentation](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions) and it said:

>  Most exceptions are defined with names that end in “Error,” similar to the naming of the standard exceptions."

So I said, alright...I'll add the Error. Happy to remove if that's OK with The Python Way (TM) :P

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: I didn't know `os.environ` was a dictionary, I thought some kind of class you had to access via `get()`. I'll change to use `[]` so it blows up if the environment variable is not set for whatever reason.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+    """Raised when trying to publish to a non-existent SNS topic"""
+    pass
+
+
+# e.g. "arn:aws:sns:eu-west-1:1234:dev_gh_team_events"
+def _topic_arn(gh_event):
+    return "{}:{}_gh_{}_events".format(
+        SNS_ARN,
+        STAGE,
+        gh_event,
```
xoen: I like that, I didn't know you could do it.

Regarding the GH acronym is quite common and short and I thought in this context (user = developer) it was appropriate to use it. We can change everything to use the full `github` name if you think it's better.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+sns = boto3.client("sns")
+
+
+class SNSTopicNotFoundError(Exception):
+    """Raised when trying to publish to a non-existent SNS topic"""
+    pass
+
+
+# e.g. "arn:aws:sns:eu-west-1:1234:dev_gh_team_events"
+def _topic_arn(gh_event):
```
xoen: My reasoning for this was to have a clear, visual distinction between what's exposed directly outside and what's only used internally by the "main" handler function.

I wonder if the lambda function entry point should be be `__main__` or AWS would sniff at it, I'll have a look.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+
+sns = boto3.client("sns")
+
+
+class SNSTopicNotFoundError(Exception):
+    """Raised when trying to publish to a non-existent SNS topic"""
+    pass
+
+
+# e.g. "arn:aws:sns:eu-west-1:1234:dev_gh_team_events"
```
xoen: Changed 👍

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: I didn't want to risk to use a keyword by accident 🗡 - I will change

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: I guess make sense. I'll change.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: Isn't Django python :trollface: ? I'll just add a link to the official GH documentation.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: 👍

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: `publish_to_sns`?

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
+import hmac
```
xoen: So module called => `github_webhooks`?

What about the entrypoint? Should we use `__main__` as entry point? In the AWS example they called the entrypoint `lambda_handler`.

I personally don't mind. The reason why I called it index/handler as that was the default in AWS console and I though was some kind of non-written convention, but if you have something in mind let me know.

(But I agree with you that the module name is not the best, probably `github_webhooks` is better)

infra/terraform/modules/data_access/gh_webhook_handler/index.py
```diff
@@ -0,0 +1,91 @@
...
+
+SNS_ARN = os.environ.get("SNS_ARN")
+STAGE = os.environ.get("STAGE")
+GH_HOOK_SECRET = os.environ.get("GH_HOOK_SECRET")
+
+sns = boto3.client("sns")
+
+
+class SNSTopicNotFoundError(Exception):
+    """Raised when trying to publish to a non-existent SNS topic"""
```
xoen: 👍

xoen:
@andyhd not familiar with Python and not sure about the testing of this, what would you do?

xoen:
@andyhd I may have a look at what you did [in the other repository](https://github.com/ministryofjustice/analytics-platform-jenkins-lib/tree/master/tests) but I may need your help 🥇

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: 👍

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: Thinking again about it, the reason why you usually see this "handler" names in the AWS Lambda examples, etc...is because this is the proper name of these things in [AWS Lambda](http://docs.aws.amazon.com/lambda/latest/dg/python-programming-model-handler-types.html).

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: The Python documentation is wrong in this case, I think. `NotFound` already conveys the fact that it is an error. They are referring to something like `SNSTopicError`, for example.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: Double underscores do have meaning in Python, so `__main__` wouldn't be a good name. I get why you put underscores on them, but in this case it's not necessary. If you want to hide module level functions and variables from code that imports the module you need to define a list named `__all__`, which contains the names of the functions and variables that should be visible.

andyhd: (But I wouldn't bother in this case)

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: It's not a big deal, obviously. If topic names need to be kept short due to constraints, then I'd say fine. It's just when I read `gh` I have to think for a second to remember it means `github` - it breaks up the reading flow. I think it's always better to be explicit by default.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: Good point. Fortunately, `error` is safe :) But `err` would be ok, I guess.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: I just meant, I followed the link and I couldn't really see what you wanted me to read - did you take the code from that page, or does the blog post highlight a particular technique or issue?

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: Yes, I take your point. But "handler" is just what AWS call a generic handler function. It's like calling the function "function". Unless you *have to* call it "handler" because it won't work otherwise, I think it's better to call it something like `publish_to_sns` because then it's self-documenting.

andyhd:
Sorry, that comment about testing was just a note to the team generally - that's why I didn't make it a requested change. Maybe @kerin thinks we don't need to write tests or has some other idea?

kerin:
No, testing is a good thing - for some aspects of the project it's unclear how we'd go about testing, but in this case I assume standard python unit tests would cover it?

infra/terraform/modules/data_access/gh_webhook_handler/index.py
andyhd: Yes, `github_webhooks` is a better name for the module. I'm not sure about `index.py` - does the file structure need to look like this? Could you have the code in `infra/terraform/modules/data_access/github_webhooks.py`? Or if the current file structure is needed, then maybe the `index.py` file should be called `publish_to_sns.py` and the `handler` function can keep its name?

xoen:
I have no idea how tests in Python works but I'll have a look at it and hopefully we'll get some kind of testing for this :)

kerin:
Are doctests still a thing @andyhd ?

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: It's not a constraint, I tried to not have very long names for other resource names which will include the lambda function name, e.g. `${var.env}_gh_webhook_handler_role_policy`. But I tend to agree with you and be explicit. In theory the limit for resource names in AWS is 256 chars (which I doubt we'll reach).

I suppose 4 chars more would be the end of the world so I'll see if I can change everything to use the full `github` name.

infra/terraform/modules/data_access/gh_webhook_handler/index.py
xoen: The beauty of calling this method `handler` is that who opens this file see immediately that `handler` is the main one and the others are utility methods. Not a specific name but very unsurprising. For example you may go to a different lambda function doing something different and because it has an `handler` method you know where to start.

But I'll change to `publish_to_sns()` 👍

andyhd: Also a good point. You've persuaded me. Keep it as `handler` 👍

andyhd:
I'm not really a fan of doctests.

andyhd:
I'm quite keen on `pytest` and the style that I've been developing that you can see in [jenkins-lib](https://github.com/ministryofjustice/analytics-platform-jenkins-lib/tree/master/tests).

xoen:
FYI: @kerin / @andyhd I made the requested changes: https://github.com/ministryofjustice/analytics-platform-ops/pull/45/commits/b23fcec67d6d7deb0cd8c10156e5536165d73c53

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+############
+#  Config  #
+############
+
+
+@pytest.fixture
+def given_the_env_is_set(sns_arn, secret):
+    os.environ["SNS_ARN"] = sns_arn
+    os.environ["STAGE"] = "test"
+    os.environ["GH_HOOK_SECRET"] = secret
```
andyhd: This will set environment variables when the tests are run. I'm sure it's not a problem, but tests should really be idempotent.

Later tests might depend on the environment being unchanged (they shouldn't of course), so it's best practice not to change the state of the system.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@pytest.fixture
+def sns_arn():
+    return "arn:aws:sns:eu-west-1:123456789012"
```
andyhd: This could just be a constant, eg: `TEST_SNS_ARN=arn:aws:sns...`

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@pytest.fixture
+def secret():
+    return "The secret shared with GitHub"
```
andyhd: This could also be a constant

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+    sns.create_topic(Name=topic)
+
+
+@pytest.fixture
+def given_the_topic_doesnt_exist(topic_arn):
+    sns = boto3.client("sns")
+    topics = sns.list_topics()["Topics"]
+
+    for t in topics:
+        if t["TopicArn"] == topic_arn:
```
andyhd: You have control over all the variables that go into making `topic_arn`, so you could replace this with a constant. Isn't this always going to be `arn:aws:sns:eu-west-1:123456789012:test_github_team_events`?

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+######################
+#  SNS setup / Moto  #
+######################
+
+
+@pytest.yield_fixture
+def given_sns_is_available():
+    with mock_sns():
+        yield
```
andyhd: I was wrong about naming this one. You were right to want to call it `mock_sns` (although that would clash with `mock_sns` from `moto`, so something similar). Then it should be passed to the following fixtures and the test as a parameter.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+@pytest.fixture
+def given_invalid_signature(state):
+    state["event"]["headers"]["X-Hub-Signature"] = "sha3=invalid"
+
+
+@pytest.fixture
+def given_valid_signature(state, secret):
+    message = "hello this is dog"
+    signature = sha1_signature(message, secret)
```
andyhd: There's no need to compute this signature every run - if the message remains constant, you can precompute the signature and hardcode it here. It's a minor optimization from a performance point of view, but it removes a bunch of code, which is always a good thing.

infra/terraform/modules/data_access/github_webhooks_handler/tests/test_publish_to_sns.py
```diff
@@ -0,0 +1,65 @@
+import json
+from moto import mock_sns
```
andyhd: This import is no longer used

andyhd: Also, there should be a blank line between `import json` and `import pytest`

infra/terraform/modules/data_access/github_webhooks_handler/tests/test_publish_to_sns.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.usefixtures(
+    "given_the_env_is_set",
+    "given_invalid_signature",
+)
+def test_cannot_publish_to_sns(event):
+    '''
+    when the signature is not valid
+    it responds {401, "Invalid signature"}
+    '''
```
andyhd: This doc-comment just repeats what the code says, so it's not needed

infra/terraform/modules/data_access/github_webhooks_handler/tests/test_publish_to_sns.py
```diff
@@ -0,0 +1,65 @@
...
+    "given_valid_signature",
+    "given_sns_is_available",
+    "given_the_topic_exists",
+)
+def test_publish_success(event):
+    '''
+    when the signature is valid
+    and the topic exists
+    it responds {201, "Event published to SNS"}
+    '''
```
andyhd: This doc-comment doesn't add anything to the code

infra/terraform/modules/data_access/github_webhooks_handler/tests/test_publish_to_sns.py
```diff
@@ -0,0 +1,65 @@
...
+    "given_valid_signature",
+    "given_sns_is_available",
+    "given_the_topic_doesnt_exist",
+)
+def test_nonexistent_topic(event, topic_arn):
+    '''
+    when the signature is valid
+    but the topic doesn't exist
+    it responds {404, "SNS topic not found"}
+    '''
```
andyhd: This comment doesn't add any information

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
+from hashlib import sha1
+from moto import mock_sns
+import boto3
+import hmac
+import json
+import os
+
+import pytest
```
xoen: So `from moto import mock_sns` is before `import pytest` because `moto` is less than `pytest` right?  Or it sorted by the name imported (e.g. `mock_sns` in this case).

Anyway, I'll sort 👍

xoen: Actually now that I commented I realised that `hashlib` is less than `hmac` so they're sorted by package name not by what you actually import.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+#################
+#  Tests state  #
+#################
```
xoen: 😱

xoen: I knew you were going to say something about these. I guess I wanted to keep things tidy, but good point about "another thing to maintain" so I'll remove them 👍

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+###########
+#  Event  #
+###########
+
+
+@pytest.fixture
+def event(state):
+    return state["event"]
```
xoen: It is actually used in the test to call the `send_to_sns(event, context)` function.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@pytest.fixture
+def event_type(event):
+    return event["headers"]["X-GitHub-Event"]
```
xoen: Good point. I'll inline this bit.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
xoen: Can do.

infra/terraform/modules/data_access/github_webhooks_handler/tests/conftest.py
xoen: Good point.

PR #46, opened 2017-05-08
Added IAM user for Auth0 SES email sending

kerin:
Adds an IAM user with SES email sending credentials, for use by Auth0 to send emails.

### DIFF
--- a/infra/terraform/global/iam.tf
+++ b/infra/terraform/global/iam.tf
@@ -1,0 +1,27 @@
+resource "aws_iam_user" "auth0_ses_user" {
+    name = "auth0-ses-user"
+}
+
+resource "aws_iam_user_policy" "auth0-ses-user" {
+    name = "auth0-ses-user"
+    user = "${aws_iam_user.auth0_ses_user.name}"
+    policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "ses:SendRawEmail",
+        "ses:SendEmail"
+      ],
+      "Effect": "Allow",
+      "Resource": "*"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_iam_access_key" "auth0_ses_user" {
+    user = "${aws_iam_user.auth0_ses_user.name}"
+}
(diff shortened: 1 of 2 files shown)


infra/terraform/global/iam.tf
```diff
@@ -0,0 +1,27 @@
+resource "aws_iam_user" "auth0_ses_user" {
+    name = "auth0-ses-user"
+}
+
+resource "aws_iam_user_policy" "auth0-ses-user" {
```
xoen: Shouldn't we call this resource `auth0_ses_user_policy` to avoid confusion?

infra/terraform/global/iam.tf
```diff
@@ -0,0 +1,27 @@
...
+      ],
+      "Effect": "Allow",
+      "Resource": "*"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_iam_access_key" "auth0_ses_user" {
```
xoen: Shouldn't we call this resource `auth0_ses_access_key` to avoid confusion?

xoen:
Not familiar with AWS SES but as far as I can tell it looks good. I would probably change the resource names to avoid confusion, feel free if you want.

infra/terraform/global/iam.tf
kerin: imo that's not necessary, as you always need to refer to Terraform resources by both type and name - always `aws_iam_user_policy.auth0-ses-user`, never just `auth0-ses-user`. That said, the names for these resources should probably be the more general `auth0-ses` rather than referring to any particular resource.

xoen: I see your point about the Terraform resource itself as you refer to it using the resource type as well.
Still, may be the case of use a more specific name for the identifier in AWS (`name = "auth0-ses-user"` bit).

As said, functionally looks good so feel free to ignore these comments on naming if you want.

kerin:
I've changed the terraform resource names to be consistent across resource types, but made the `name` identifiers more explicit so they make more sense in AWS.

xoen:
@kerin 👍

PR #47, opened 2017-05-09
Enable deletion of centrally stored logfiles after 30 days

kerin:
## What

1. Downgrades AWS Elasticsearch to v2.3 due to [v5.1 incompatibility with Curator](https://github.com/elastic/curator#compatibility-matrix) 
2. Adds elasticsearch DNS alias in the form `elasticsearch.ENV.domain` to allow reprovisioning Elasticsearch cluster without requiring all other upstream services to update their ES hostname entry
3. Updates existing elasticsearch hostname entries in fluentd chart configs
4. Enables support for Kubernetes alpha CronJob resource
5. Provides Helm chart for Curator log deletion cron job

## How to review

As we are currently using a single shared dev environment this is difficult to deploy and test in isolation without rolling back changes already made to the development environment, which is particularly problematic in this case as enabling alpha CronJob support requires redeploying the entire cluster - as such there are few options other than simply reading through the changes.


xoen:
I like the simpler ES hostname. Shouldn't also change in the kibana chart's config? (`chart-env-config/alpha/kibana.yml`)

charts/elasticsearch-curator/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
xoen: Do you reckon is the case of adding a short `README.md` with the command to install the chart and maybe other details as we do in others charts (e.g. [rstudio helm chart's README.md](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/charts/rstudio/README.md))?

charts/elasticsearch-curator/templates/cronjob.yml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  schedule: "{{ .Values.schedule }}"
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          labels:
+            app: {{ template "fullname" . }}
+          annotations:
+            checksum/secrets: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
```
xoen: This is still checksumming the `secrets.yaml` before the values are interpolated right? I suppose it's better than nothing.

It's something to keep in mind, change just a value in the secrets and upgrading the helm chart would not re-create the pod with the new environment variables with the correct values unfortunately.

charts/elasticsearch-curator/values.yaml
```diff
@@ -0,0 +1,10 @@
+image:
+  repository: bobrik/curator
+  tag: 3.5.1
+  pullPolicy: IfNotPresent
+schedule: "0 1 * * *"
```
xoen: Even if I roughly know how the cron fields work they always puzzle me. I wonder if a comment would help, e.g. "Run every day at 1:00am". Or just ignore me.

charts/elasticsearch-curator/templates/cronjob.yml
kerin: Actually the other day we realised that this checksums the _rendered_ file after all, with values interpolated.

kerin:
@xoen I've addressed your comments - could you take another look?

charts/elasticsearch-curator/templates/cronjob.yml
xoen: That's great news :)

xoen:
@kerin looks great 👍

PR #49, opened 2017-05-09
Enable adding multiple users at once

andyhd:
## What

* Move build parameters into Jenkinsfile
* Call updated python script with list of email addresses

## How to review

1. Run the `grant-access-to-app` Control Panel Jenkins job
2. See the parameters are now "APP_NAME" and "EMAILS"
3. Enter multiple email addresses separated by semicolons (`;`)
4. See the job run
5. See users with the given email addresses have been granted access to the specified app

See [Enable user to provide semicolon delimited list of email addresses as part of the Jenkins grant-access job](https://trello.com/c/L5cA87Mn)

### DIFF
--- a/jenkinsfiles/grant_access_to_app.groovy
+++ b/jenkinsfiles/grant_access_to_app.groovy
@@ -20,17 +20,17 @@ node {
     // Creates passwordless user in Auth0 and adds to app group
     // NOTE: Groups provided by Auth0 Authorization Extension
     stage ("Create Auth0 passwordless user and add to app group") {
+      env.AUTHZ_API_ID = 'urn:auth0-authz-api'
       withCredentials([
-        usernamePassword(
-            credentialsId: 'auth0-mgmt-api',
-            usernameVariable: 'CLIENT_ID',
-            passwordVariable: 'CLIENT_SECRET'),
         string(
-            credentialsId: 'authz_api_url',
+            credentialsId: 'AUTH0_CLIENT_ID',
+            variable: 'CLIENT_ID'),
+        string(
+            credentialsId: 'AUTH0_CLIENT_SECRET',
+            variable: 'CLIENT_SECRET'),
+        string(
+            credentialsId: 'AUTHZ_API',
             variable: 'AUTHZ_API_URL'),
-        string(
-            credentialsId: 'authz_api_id',
-            variable: 'AUTHZ_API_ID'),
         string(
             credentialsId: 'AUTH0_DOMAIN',
             variable: 'AUTH0_DOMAIN')
--- a/jenkinsfiles/grant_access_to_app.groovy
+++ b/jenkinsfiles/grant_access_to_app.groovy
@@ -1,5 +1,18 @@
 node {
 
+    properties([
+        parameters([
+            string(
+                name: 'APP_NAME',
+                description: 'Shiny App name',
+                defaultValue: ''),
+            string(
+                name: 'EMAILS',
+                description: 'Semicolon delimited list of email addresses',
+                defaultValue: ''),
+        ])
+    ])
+
     stage ("Git checkout") {
         checkout scm
     }
@@ -22,7 +35,7 @@ node {
             credentialsId: 'AUTH0_DOMAIN',
             variable: 'AUTH0_DOMAIN')
       ]) {
-        sh "/usr/local/bin/grant_access ${env.APP_NAME} ${env.EMAIL}"
+        sh "/usr/local/bin/grant_access ${env.APP_NAME} ${env.EMAILS}"
       }
     }
 }


jenkinsfiles/grant_access_to_app.groovy
```diff
@@ -1,28 +1,41 @@
...
+      env.AUTHZ_API_ID = 'urn:auth0-authz-api'
       withCredentials([
-        usernamePassword(
-            credentialsId: 'auth0-mgmt-api',
-            usernameVariable: 'CLIENT_ID',
-            passwordVariable: 'CLIENT_SECRET'),
         string(
-            credentialsId: 'authz_api_url',
-            variable: 'AUTHZ_API_URL'),
+            credentialsId: 'AUTH0_CLIENT_ID',
```
xoen: Shouldn't we convert this to use `usernamePassword` as we do [in the other jenkins job](https://github.com/ministryofjustice/analytics-platform-ops/blob/2a50d01359b328d3f45860c0e5961302fb46c327/jenkinsfiles/grant_access_to_app.groovy#L11)?

jenkinsfiles/grant_access_to_app.groovy
```diff
@@ -1,28 +1,41 @@
...
+    ])
+
     stage ("Git checkout") {
         checkout scm
     }
 
     // Creates passwordless user in Auth0 and adds to app group
     // NOTE: Groups provided by Auth0 Authorization Extension
     stage ("Create Auth0 passwordless user and add to app group") {
+      env.AUTHZ_API_ID = 'urn:auth0-authz-api'
```
xoen: Is this is a constant, shouldn't it be hidden in the `grant_access` script?

jenkinsfiles/grant_access_to_app.groovy
```diff
@@ -1,28 +1,41 @@
...
 
+    properties([
+        parameters([
+            string(
+                name: 'APP_NAME',
+                description: 'Shiny App name',
+                defaultValue: ''),
+            string(
+                name: 'EMAILS',
+                description: 'Semicolon delimited list of email addresses',
```
xoen: Just out of curiosity, why did you use semicolons (`;`) instead of commas (`,`) as email delimiters?

xoen:
Added a couple of comments @andyhd but otherwise looks good to me.

jenkinsfiles/grant_access_to_app.groovy
andyhd: The credential Kinds are actually "Secret Text" (in Jenkins > Credentials UI), which translates to `string` in the Jenkinsfile. If the credentials Kind was "Username with password", then I could use `usernamePassword` here.

jenkinsfiles/grant_access_to_app.groovy
andyhd: Maybe - it doesn't seem very sensitive. I'll add it as a credential instead.

jenkinsfiles/grant_access_to_app.groovy
andyhd: That's what was requested in the Trello ticket: https://trello.com/c/L5cA87Mn

jenkinsfiles/grant_access_to_app.groovy
```diff
@@ -22,7 +35,7 @@ node {
             credentialsId: 'AUTH0_DOMAIN',
             variable: 'AUTH0_DOMAIN')
       ]) {
-        sh "/usr/local/bin/grant_access ${env.APP_NAME} ${env.EMAIL}"
+        sh "/usr/local/bin/grant_access ${env.APP_NAME} ${env.EMAILS}"
```
xoen: Ahaha how did I missed this meme? (rhetoric question, no need to reply :P )

jenkinsfiles/grant_access_to_app.groovy
andyhd: I'll delimit on `;`, `,` or ` `(space) as discussed

jenkinsfiles/grant_access_to_app.groovy
xoen: Feel free to change if you want. The only reason I used the more generic "Secret Text" was because I struggled to find documentation for the jenkins credentials plugin. If we think CLIENT_ID/CLIENT_SECRET as username/password so maybe the `usernamePassword` credential type may be more appropriate. If.

jenkinsfiles/grant_access_to_app.groovy
```diff
@@ -1,5 +1,18 @@
...
 
+    properties([
+        parameters([
+            string(
+                name: 'APP_NAME',
+                description: 'Shiny App name',
+                defaultValue: ''),
+            string(
+                name: 'EMAILS',
+                description: 'Semicolon delimited list of email addresses',
```
andyhd: Will split on `,`, `;` or space, as discussed

PR #52, opened 2017-05-10
Terraform alpha env: Pass `dns_zone_id` to `logging_elasticsearch` module

xoen:
## What

Tried to run `$ terraform plan` in the `alpha` environment but it complained about the missing `dns_zone_id` variable in the `logging_elasticsearch` module (as part of [Kerin's work on ES](https://github.com/ministryofjustice/analytics-platform-ops/pull/47))

## How to review

1. Run `$ terraform plan`
2. Does it return an error?
3. If not this fixes the problem


xoen:
Checked with Kerin offline.

PR #55, opened 2017-05-10
GH Webhooks handler responds 202 ACCEPTED as it should.

xoen:
## What

The GH Webhooks handler don't process the GH events directly but rather submit them to AWS SNS for later processing.
So respond with a `202 ACCEPTED` is more appropriate.

## Why

See GOV.UK's Style [Guidelines on API](https://github.com/alphagov/styleguides/blob/master/api.md):

> Use 202 for times when the data is valid, but we've queued
> the persistence for background processing, as described in
> RFC 7231:

> "The request has been accepted for processing, but the
> processing has not been completed.


xoen:
Thanks @andyhd

PR #57, opened 2017-05-12
API Gateway resources to access to POST /github endpoint

xoen:
## What

Terraform resources to deploy the API Gateway to handle the GitHub webhooks.

## Why

To be able to deploy the API very quickly without having to click around in the AWS Console (which is also error prone)

## Ticket

https://trello.com/c/YkoHJIMD/113-create-aws-api-gateway-with-terraform

## How to review

1. Delete the `dev_analytics` API [from the AWS Console](https://eu-west-1.console.aws.amazon.com/apigateway/home?region=eu-west-1#/apis)
2. Run `$ terraform plan/apply -target=module.data_access` from within `/infra/terraform/environments/dev`
3. Go to the [GitHub webhooks page](https://github.com/organizations/moj-analytical-services/settings/hooks/13614174#delivery-response), re-deliver one of the last events 
4. You should get back a `202 ACCEPTED` or a `404 NOT FOUND`, not a `500 INTERNAL SERVER ERROR` 

## About environment isolation

We decided to have an API Gateway per environment (called
`${ENV}_analytics`) with a stage called after the environment
(e.g. `dev`, `alpha`, etc...).
The stage will became part of the API URL, e.g. [https://abc123.execute-api.eu-west-1.amazonaws.com/dev](https://abc123.execute-api.eu-west-1.amazonaws.com/dev)
which is very unsurprising as you can always see the environment you're
calling.

## POST /github endpoint

At the moment the API is composed only by the following endpoint:

 - `POST /github`
   - Publishes the GitHub event to SNS and returns `202 ACCEPTED`
   - If the SNS doesn't exist (because we're not interested in handle it),
   it returnd `404 NOT FOUND`
   - If the secret provided by GitHub (via `X-Hub-Signature` header) is
   invalid returns `401 UNAUTHORIZED`

## Terraform details

We decided to have 3 modules to avoid dump everything in a single module:
- `webhooks_api` contains the API Gateway resources
- `notifications` contains the SNS topics and the lambda function
   which publishes the GitHub events to these topics
- `data_access` will contain (currently empty) the lambda functions
   to handle these events and add/delete IAM resources

In AWS API GateWay parlance these are the terraform resources created:
 - `*_gateway_rest_api` for the whole API
 - `*_deployment` the deployment which makes the API methods available from
 the outside. APIs are deployed to a "stage"
 - `*_resource` is the resource which can manipulated via its methods,
 `/github` in our case
 - `*_method` is the method to manipulate a specific resource,
it's the combination of a HTTP verb and the resource name, e.g. `POST /github`.
Note this method **does not requires** authentication because the Lambda
function will check the GitHub webhook signature
 - `*_integration` is the bit which converts the request format
 to the "backend" endpoint's format. In our case we just use the Lambda Proxy
 integration, which means the request goes straight to the Lambda function
 without alterations
 - `*_role`, the IAM role assumed by the API when calling the Lambda function
 - `*_role_policy`, the IAM policy which grants the role the permission to
 invoke the Lambda function


infra/terraform/environments/alpha/main.tf
```diff
@@ -86,6 +86,8 @@ module "container_registry" {
 
 module "data_access" {
```
kerin: I'm not sure `data_access` is the best name for this module, as that's not really what it's doing - 'data access' is really the shorthand name of our sprint goal. I think `webhook_api` would probably be better?

infra/terraform/environments/alpha/main.tf
```diff
@@ -86,6 +86,8 @@ module "container_registry" {
...
     source = "../../modules/data_access"
+
+    region = "${var.region}"
     env = "${var.env}"
     sns_arn = "${var.sns_arn}"
```
kerin: It looks like `sns_arn` isn't actually used anywhere, so it should be removed.

infra/terraform/environments/alpha/main.tf
```diff
@@ -86,6 +86,8 @@ module "container_registry" {
...
     gh_hook_secret = "${var.gh_hook_secret}"
```
kerin: This secret will have to be populated in the github webhook config, so it should be defined as a Terraform output value so it can be easily extracted.

infra/terraform/modules/data_access/api_gateway.tf
```diff
@@ -0,0 +1,77 @@
+resource "aws_api_gateway_rest_api" "analytics" {
+  name = "${var.env}_analytics"
+}
+
+# API Gateway stage (to give access to API from the outside)
+resource "aws_api_gateway_deployment" "analytics" {
+  rest_api_id = "${aws_api_gateway_rest_api.analytics.id}"
+  stage_name  = "${var.env}"
```
kerin: I'm not convinced the API stage should be named after the environment - it feels like it should be named `production` or even `api` - that way the stage names are the same in all environments, and represent their role in that environment specifically, rather than the environment they're part of.

infra/terraform/environments/alpha/main.tf
xoen: Thanks for your review @kerin! I see what you mean. Not sure although.

It's still not 100% clear how I'll structure the code (and I'll probably move few things down the line as I go) but I was planning to have all the resources related to the data access in this module, e.g. also the lambda functions to perform atomic operations like, `create_iam_role`, `delete_iam_role`, etc...which may make sense also outside a `webhook_api` module and which we may manually trigger and/or from an UI later at some point.

infra/terraform/environments/alpha/main.tf
xoen: Weird, I can't remember, let me double check...

xoen: So, it was added in another PR but it's used by the [lambda function](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/lambda.tf#L22) to build the [SNS topic where GH events are published to](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/github_webhooks_handler/github_webhooks.py#L49).

infra/terraform/environments/alpha/main.tf
xoen: I though about that at some point actually, but then I thought it wasn't really needed as it's relatively easy to get from the config.

I'll add an output variable 👍

infra/terraform/modules/data_access/api_gateway.tf
xoen: As the stage becomes the last part of the API URL (e.g. `https://rand_stuff.execute-api.eu-west-1.amazonaws.com/dev/github`) it's really clear which environment will end up handling the webhook which to me is a benefit as it avoid surprises/confusion.

But, it's easy to change and I can if you really think it's better to have something like `https://rand_stuff.execute-api.eu-west-1.amazonaws.com/api/github` :)

xoen:
@kerin I think it's better now. I did the amendments requested and split things in 3 terraform modules. I like it more now.

The `data_access` module is now empty until we start to work on the lambda functions which handle the GitHub events.

PR #62, opened 2017-05-23
Handle GitHub "organization" events to add/remove IAM roles in AWS

xoen:
## What

This is to handle the organization GitHub events arriving from the corresponding SNS topic (`${ENV}_github_organization_events`)

These events are handled by the `organization_events` "controller" lambda function which then invoke the `create_user_role`/`delete_user_role` asyncrounously.

**NOTE**: The `delete_user_role` lambda function has to detach all the role's policies before being able to delete it.

## Ticket:

Part of ticket: https://trello.com/c/UZ8qooqO/116-xl-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create-update-delete-aws-resources

## TODO / Ideas

- [x] Test the "controller" function
- [x] Test better the users' lambda functions?

### DIFF
--- a/infra/terraform/modules/data_access/lambda_organization_events.tf
+++ b/infra/terraform/modules/data_access/lambda_organization_events.tf
@@ -1,0 +1,97 @@
+# Zip the lambda function before the actual deploy
+data "archive_file" "organization_events_zip" {
+    type        = "zip"
+    source_dir  = "${path.module}/organization_events"
+    output_path = "/tmp/organization_events.zip"
+}
+
+# Lambda function which handles GH organization events
+resource "aws_lambda_function" "organization_events" {
+    description = "Updates IAM resources in response to GH organization events"
+    filename = "/tmp/organization_events.zip"
+    source_code_hash = "${data.archive_file.organization_events_zip.output_base64sha256}"
+    function_name = "${var.env}_organization_events"
+    role = "${aws_iam_role.organization_events_role.arn}"
+    handler = "organization_events.event_received"
+    runtime = "python3.6"
+    timeout = 10
+    depends_on = ["data.archive_file.organization_events_zip"]
+    environment {
+        variables = {
+            CREATE_ROLE_ARN = "${aws_lambda_function.create_user_role.arn}",
+            DELETE_ROLE_ARN = "${aws_lambda_function.delete_user_role.arn}",
+        }
+    }
+}
+
+# Lambda function triggered by SNS notification
+resource "aws_sns_topic_subscription" "organization_events" {
+    topic_arn = "${var.organization_events_topic_arn}"
+    protocol = "lambda"
+    endpoint = "${aws_lambda_function.organization_events.arn}"
+}
+
+# Role running the lambda function
+resource "aws_iam_role" "organization_events_role" {
+    name = "${var.env}_organization_events_role"
+    assume_role_policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": "sts:AssumeRole",
+      "Principal": {
+        "Service": "lambda.amazonaws.com"
+      },
+      "Effect": "Allow"
+    }
+  ]
+}
+EOF
+}
+
+# Policies for the 'organization_events_role' role
+resource "aws_iam_role_policy" "organization_events_role_policy" {
+    name = "${var.env}_organization_events_role_policy"
+    role = "${aws_iam_role.organization_events_role.id}"
+    policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "CanInvokeLambdaFunctions",
+      "Effect": "Allow",
+      "Action": [
+        "lambda:InvokeFunction"
+      ],
+      "Resource": [
+        "${aws_lambda_function.create_user_role.arn}",
+        "${aws_lambda_function.delete_user_role.arn}"
+      ]
+    },
+    {
+      "Sid": "CanLog",
+      "Effect": "Allow",
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:PutLogEvents",
+        "logs:DescribeLogStreams"
+      ],
+      "Resource": [
+        "arn:aws:logs:*:*:*"
+      ]
+    }
+  ]
+}
+EOF
+}
+
+# Permission to invoke the lambda function
+resource "aws_lambda_permission" "allow_organization_events_invocation" {
+    statement_id = "AllowExecutionFromSNS"
+    action = "lambda:InvokeFunction"
+    function_name = "${aws_lambda_function.organization_events.arn}"
+    principal = "sns.amazonaws.com"
+    source_arn = "${var.organization_events_topic_arn}"
+}
--- a/infra/terraform/modules/data_access/lambda_organization_events.tf
+++ b/infra/terraform/modules/data_access/lambda_organization_events.tf
@@ -31,6 +31,15 @@ resource "aws_sns_topic_subscription" "organization_events" {
     endpoint = "${aws_lambda_function.organization_events.arn}"
 }
 
+# Permission to invoke the lambda function
+resource "aws_lambda_permission" "allow_organization_events_invocation" {
+    statement_id = "AllowExecutionFromSNS"
+    action = "lambda:InvokeFunction"
+    function_name = "${aws_lambda_function.organization_events.arn}"
+    principal = "sns.amazonaws.com"
+    source_arn = "${var.organization_events_topic_arn}"
+}
+
 # Role running the lambda function
 resource "aws_iam_role" "organization_events_role" {
     name = "${var.env}_organization_events_role"
@@ -85,13 +94,4 @@ resource "aws_iam_role_policy" "organization_events_role_policy" {
   ]
 }
 EOF
-}
-
-# Permission to invoke the lambda function
-resource "aws_lambda_permission" "allow_organization_events_invocation" {
-    statement_id = "AllowExecutionFromSNS"
-    action = "lambda:InvokeFunction"
-    function_name = "${aws_lambda_function.organization_events.arn}"
-    principal = "sns.amazonaws.com"
-    source_arn = "${var.organization_events_topic_arn}"
 }
--- a/infra/terraform/modules/data_access/lambda_organization_events.tf
+++ b/infra/terraform/modules/data_access/lambda_organization_events.tf
@@ -43,20 +43,7 @@ resource "aws_lambda_permission" "allow_organization_events_invocation" {
 # Role running the lambda function
 resource "aws_iam_role" "organization_events_role" {
     name = "${var.env}_organization_events_role"
-    assume_role_policy = <<EOF
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Action": "sts:AssumeRole",
-      "Principal": {
-        "Service": "lambda.amazonaws.com"
-      },
-      "Effect": "Allow"
-    }
-  ]
-}
-EOF
+    assume_role_policy = "${data.aws_iam_policy_document.lambda_assume_role.json}"
 }
 
 # Policies for the 'organization_events_role' role
--- a/infra/terraform/modules/data_access/lambda_users.tf
+++ b/infra/terraform/modules/data_access/lambda_users.tf
@@ -18,8 +18,8 @@ resource "aws_lambda_function" "create_user_role" {
     depends_on = ["data.archive_file.users_zip"]
     environment {
         variables = {
-            ACCOUNT_ID = "${var.account_id}",
             STAGE = "${var.env}",
+            SAML_PROVIDER_ARN = "arn:aws:iam::${var.account_id}:saml-provider/auth0",
         }
     }
 }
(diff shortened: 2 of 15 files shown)


infra/terraform/modules/data_access/lambda_users.tf
```diff
@@ -0,0 +1,155 @@
...
+    function_name = "${var.env}_create_user_role"
+    role = "${aws_iam_role.create_user_role.arn}"
+    handler = "users.create_user_role"
+    runtime = "python3.6"
+    timeout = 10
+    depends_on = ["data.archive_file.users_zip"]
+    environment {
+        variables = {
+            STAGE = "${var.env}",
+            SAML_PROVIDER_ARN = "arn:aws:iam::${var.account_id}:saml-provider/auth0",
```
kerin: Just as a note - once I've merged in my federated identity Terraform resources this should change to a Terraform `${resource.name.arn}` reference instead.

infra/terraform/modules/data_access/lambda_users.tf
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+}
+
+# Role assumed by the 'create_user_role' lambda function
+resource "aws_iam_role" "create_user_role" {
```
kerin: This role seems the same as the `organization_events_role` - does it make sense to have two separate roles rather than a `lambda_function` role?

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+LAMBDA_ARNS = {
+    "create_user_role": os.environ["CREATE_ROLE_ARN"],
+    "delete_user_role": os.environ["DELETE_ROLE_ARN"],
+}
+
+
+def event_received(sns_event, context):
+    print("SNS event received = {}".format(json.dumps(sns_event)))
```
kerin: It'd be better to use the `logging` module rather than `print`, so we can use log levels.

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,54 @@
...
+        function=LAMBDA_ARNS["delete_user_role"],
+        payload=payload(event)
+    )
+
+
+def invoke_lambda(function, payload):
+    client = boto3.client("lambda")
+    client.invoke(
+        FunctionName=function,
+        Payload=bytes(json.dumps(payload), "utf8"),
```
kerin: Why the use of `bytes` here?

infra/terraform/modules/data_access/lambda_users.tf
xoen: Good point and I already thought about it ;)

infra/terraform/modules/data_access/organization_events/organization_events.py
xoen: It's the way [boto3's invoke()](https://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.invoke) works. I tried to simply pass a dictionary but it didn't work: The documentation says:

> Payload (bytes or seekable file-like object) -- JSON that you want to provide to your Lambda function as input.

Hence the `bytes` call to convert the string into the format required by boto3 (I agree I would have preferred the API to be a bit higher level and for them to call `bytes` internally on their end).

infra/terraform/modules/data_access/organization_events/organization_events.py
xoen: I tried to keep it simple but I can change if you want.

infra/terraform/modules/data_access/lambda_users.tf
xoen: I did this way so that for example the `create_role` lambda function doesn't have the policy to perform `iam:DeleteRole`.

I guess these two lambda function are closely related so we may go for a single role with the policies to create/delete an IAM role and also detach policies, which is a bit simpler.

infra/terraform/modules/data_access/organization_events/organization_events.py
kerin: the [logging module](http://docs.aws.amazon.com/lambda/latest/dg/python-logging.html) is pretty simple, so I'd say it's worth using.

infra/terraform/modules/data_access/organization_events/organization_events.py
kerin: fair enough!

infra/terraform/modules/data_access/lambda_users.tf
kerin: Sorry, misunderstood what I was looking at there. Separate roles makes sense if they have different permission levels - looks like the `assume_role_policy` could be defined once for all roles though?

xoen:
@andyhd I added some tests for the other lambda function and made the little change suggested earlier. I'm going to use the python logger instead of print as suggested by Kerin but you can have a look if you want.

xoen:
@kerin FYI: Changed the code to use the Python logger as suggested: https://github.com/ministryofjustice/analytics-platform-ops/pull/62/commits/996f75800ffea5fe0bada623a28cf622e86e9f74

infra/terraform/modules/data_access/lambda_users.tf
xoen: Thanks for the chat yesterday @kerin - Didn't know/remember there was a terraform resource for the policy. I'll give it a go 👍

xoen: Factored out in this commit: https://github.com/ministryofjustice/analytics-platform-ops/pull/62/commits/9d258eeb09150ffd8c17bf5134df1f66358d2694

infra/terraform/modules/data_access/users/users.py
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+        ]
+    }
+
+    client = boto3.client("iam")
+    client.create_role(
+        RoleName=role_name(event["username"]),
+        Path="/users/",
+        AssumeRolePolicyDocument=json.dumps(trust_relationship),
```
andyhd: Does this `json.dumps` not need to be encoded to utf8 bytes?

infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
```diff
@@ -0,0 +1,35 @@
...
+import pytest
+
+import organization_events
+
+
+@pytest.mark.usefixtures(
+    "given_the_env_is_set",
+    "given_lambda_is_available",
+)
+def test_event_received_member_added(lambda_client_mock, member_added_event, create_role_arn, payload_bytes):
```
andyhd: The name of this test (and the others) doesn't follow the BDD style. It should read more like a natural sentence, eg: `test_when_event_received_a_member_is_added`

infra/terraform/modules/data_access/organization_events/organization_events.py
xoen: Not familiar with Python sorry, I'll give it a go.

I didn't know strings have this method, would the returned value compatibles with Boto's Payload param (bytes or seekable file-like object)?

PS: It ASCII encoding is already UTF8 as ASCII is a subset of UTF8 which is retro-compatible with it.
UTF8 is a variable length encoding but the encoding for the most common chars, e.g. ASCII chars, use less bytes (1 in the case of ASCII) and less common chars use 2-4 bytes.

infra/terraform/modules/data_access/lambda_organization_events.tf
```diff
@@ -0,0 +1,84 @@
...
+    function_name = "${var.env}_organization_events"
+    role = "${aws_iam_role.organization_events_role.arn}"
+    handler = "organization_events.event_received"
+    runtime = "python3.6"
+    timeout = 10
+    depends_on = ["data.archive_file.organization_events_zip"]
+    environment {
+        variables = {
+            CREATE_ROLE_ARN = "${aws_lambda_function.create_user_role.arn}",
+            DELETE_ROLE_ARN = "${aws_lambda_function.delete_user_role.arn}",
```
xoen: I kept these names sweet and short but I get these may be confused for "roles", not "lambdas"'s  ARNs so I'll change.

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+'''
+Environment variables:
+ - CREATE_ROLE_ARN, ARN of the lambda function to create the IAM role
+ - DELETE_ROLE_ARN, ARN of the lambda function to delete the IAM role
+ - LOG_LEVEL, change the logging level (default is "DEBUG"). Must be one of
+   the python logging supported levels: "CRITICAL", "ERROR", "WARNING",
+   "INFO" or "DEBUG" (See: https://docs.python.org/2/library/logging.html#logging-levels)
+'''
```
xoen: 👍

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+LOG = logging.getLogger(__package__)
```
xoen: As discussed offline I copied from the python documentation but I will change.

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,53 @@
...
+LOG.setLevel(LOG_LEVEL)
+
+
+def event_received(sns_event, context):
+    EVENT_HANDLERS = {
+        "member_added": os.environ["CREATE_ROLE_ARN"],
+        "member_removed": os.environ["DELETE_ROLE_ARN"],
+    }
+
+    LOG.debug("SNS event received = {}".format(json.dumps(sns_event)))
```
xoen: The reason why I started to do it is because AWS CloudWatch logs shows JSON nicely with syntax highlighting, etc...which is good for debugging.

It's just a minor thing but it's nicer :)

infra/terraform/modules/data_access/organization_events/organization_events.py
```diff
@@ -0,0 +1,53 @@
...
+        FunctionName=function,
+        Payload=bytes(json.dumps(payload), "utf8"),
+        InvocationType="Event",
+    )
+
+
+def payload(event):
+    return {
+        "username": event["membership"]["user"]["login"],
+    }
```
xoen: I like to wrap things in small function especially when used 2+ times but after the refactored we use this only once so I guess I can just inline it as it's relatively simple 👍

infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
xoen: 🙏

infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
```diff
@@ -0,0 +1,35 @@
...
+)
+def test_event_received_member_removed(lambda_client_mock, member_removed_event, delete_role_arn, payload_bytes):
+    organization_events.event_received(member_removed_event, None)
+
+    # Test right lambda function is invoked asyncronously
+    lambda_client_mock.invoke.assert_called_with(
+        FunctionName=delete_role_arn,
+        Payload=payload_bytes,
+        InvocationType="Event",
+    )
```
xoen: Thanks, it's a nice trick. I managed to make it work by removing some of the fixtures and importing the constants 🤘

infra/terraform/modules/data_access/users/tests/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+TEST_SAML_PROVIDER_ARN = "arn:aws:iam::123456789012:saml-provider/auth0"
+TEST_STAGE = "test"
+TEST_USERNAME = "alice"
+
+TEST_ROLE_POLICY_ARN = "test_policy_arn"
+
+
+@pytest.fixture
+def username():
+    return TEST_USERNAME
```
xoen: Thanks for pointing out how to import these constants from the `conftest.py` file  👍 - I should probably read some Python "theory" :P, I removed some of the fixtures and simplified the tests.

infra/terraform/modules/data_access/users/users.py
```diff
@@ -0,0 +1,75 @@
...
+                },
+                "Action": "sts:AssumeRoleWithSAML",
+                "Condition": {
+                    "StringEquals": {
+                        "SAML:aud": "https://signin.aws.amazon.com/saml"
+                    }
+                }
+            }
+        ]
+    }
```
xoen: I think this is a good example of "good duplication" in tests which I was trying to explain earlier. 

The nice thing of having it duplicated in the tests is that we don't rely on a function in the tested code to be correct, e.g. if we would break the `trust_relationship()` function, the test would fail (as it should).

infra/terraform/modules/data_access/users/users.py
xoen: I missed this, I'll change.

On a related note, do you think we need to pass `"utf8"` all the time considering it's the default encoding and it's a good default? It feels like we clutter the code without much benefit (and less code is usually a good thing).

infra/terraform/modules/data_access/users/users.py
```diff
@@ -0,0 +1,75 @@
...
+    for policy in policies["AttachedPolicies"]:
+        client.detach_role_policy(
+            RoleName=role_name,
+            PolicyArn=policy["PolicyArn"],
+        )
+
+
+def role_name(username):
+    return "{env}_{username}_role".format(
+        env=os.environ["STAGE"],
```
xoen: We could do and default to `dev`. On the other end, I feel like there is some value on having this to blow up when `STAGE` is not set as it would highlight some possible wiring problem which otherwise may go unnoticed.

infra/terraform/modules/data_access/organization_events/organization_events.py
andyhd: 👍

PR #63, opened 2017-05-25
Amendments to as per Andy comments (PR #62)

xoen:
- Using `str#encode` instead of `bytes()`
- Renamed environment variables to be explicit about the fact
  these ARNs belongs to lambda functions
- Moved module documentation at the beginning of the file
- Using `getLogger(__name__)` instead of `__package__`
- Inline lambda functions payload instead of having it in a
  function
- Removed some of the tests fixtures and used pytest's parametrize
  to remove some duplication

See [PR #62: Handle GitHub "organization" events to add/remove IAM roles in AWS](https://github.com/ministryofjustice/analytics-platform-ops/pull/62#pullrequestreview-40250908)

### DIFF
--- a/infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
+++ b/infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
@@ -5,31 +5,24 @@ import pytest
 import organization_events
 
 
+from tests.conftest import sns_event, TEST_PAYLOAD_BYTES, TEST_CREATE_ROLE_ARN, \
+    TEST_DELETE_ROLE_ARN
+
+
+@pytest.mark.parametrize("event,lambda_arn", [
+    (sns_event("member_added"), TEST_CREATE_ROLE_ARN),
+    (sns_event("member_removed"), TEST_DELETE_ROLE_ARN)
+])
 @pytest.mark.usefixtures(
     "given_the_env_is_set",
     "given_lambda_is_available",
 )
-def test_event_received_member_added(lambda_client_mock, member_added_event, create_role_arn, payload_bytes):
-    organization_events.event_received(member_added_event, None)
+def test_when_event_received_is_handled(lambda_client_mock, event, lambda_arn):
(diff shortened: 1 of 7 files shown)


infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
```diff
@@ -4,32 +4,26 @@
...
-    # Test right lambda function is invoked asyncronously
-    lambda_client_mock.invoke.assert_called_with(
-        FunctionName=create_role_arn,
-        Payload=payload_bytes,
-        InvocationType="Event",
-    )
+from tests.conftest import PAYLOAD_BYTES
+from tests.conftest import sns_event
+from tests.conftest import TEST_CREATE_ROLE_ARN
+from tests.conftest import TEST_DELETE_ROLE_ARN
```
xoen: Style question: What about the order of these @andyhd ? Alphabetic order feels a bit ugly.

infra/terraform/modules/data_access/organization_events/tests/test_organization_events.py
```diff
@@ -4,32 +4,26 @@
...
 
 
+@pytest.mark.parametrize("event,lambda_arn", [
```
xoen: This is how I ended up handling the `pytest.mark.parametrize`, I tweaked the `sns_event()` function to just get the GH event name and I'm using it here. It seems nicer but have a look, you may have some feedback.

infra/terraform/modules/data_access/users/tests/test_delete_user_role.py
```diff
@@ -2,21 +2,25 @@
 
 import users
 
+from tests.conftest import ROLE_NAME
```
xoen: Should I name all these constants with the `TEST_` prefix for consistency? What do you reckon @andyhd ?

andyhd: Yep

PR #64, opened 2017-05-26
add SoftNAS Terraform module, setup SoftNAS in dev env, and create matching SSH bastion security groups

kerin:
## What

Due to unacceptable performance with the current NFS service, AWS Elastic File System, a SoftNAS terraform module (`user_nfs_softnas`) has been added alongside the existing `user_nfs` module. This has been added in parallel rather than replacing the existing module so that data can be migrated to SoftNAS before removing the EFS module.

## How to review

Follow the instructions detailed in the readme to deploy SoftNAS to the dev env, then ssh into a cluster node and mount the NFS share to confirm.


README.md
```diff
@@ -59,6 +59,11 @@ Global AWS resources (DNS and S3 buckets) only need to be created once, and are
   3. `$ terraform plan` - check that Terraform plans to create two S3 buckets (Terraform and Kops state) and a root DNS zone in Route53.
   4. `$ terraform apply` to create resources
 
+### NFS server licensing
```
xoen: 👍 for the readme

infra/terraform/global/iam.tf
```diff
@@ -25,3 +25,76 @@ EOF
...
+        "ec2:DisassociateAddress",
+
+        "s3:CreateBucket",
+        "s3:Delete*",
+        "s3:Get*",
+        "s3:List*",
+        "s3:Put*"
+      ],
+      "Resource": [
+        "*"
```
xoen: I guess there is no way to be more specific in this case. Worth bearing in mind that with this policy it's possible to do a lot of stuff (didn't read the rest and not familiar with SoftNAS, so apologies if this is actually the minimum required).

kerin: Yep, I'd rather that this policy was more specific, but SoftNAS don't give any detail on what operations they actually perform, they just publish this policy doc. At some point we can try reducing the scope of this and see what impact it has on SoftNAS.

infra/terraform/global/outputs.tf
```diff
@@ -25,3 +25,7 @@ output "auth0_ses_access_key_id" {
 output "auth0_ses_secret_key" {
     value = "${aws_iam_access_key.auth0_ses.secret}"
 }
+
+output "softnas_iam_role_arn" {
```
xoen: Do we need this?

infra/terraform/global/iam.tf
```diff
@@ -25,3 +25,76 @@ EOF
...
+        "ec2:DescribeTags",
+        "ec2:DescribeInstances",
+        "ec2:ModifyNetworkInterfaceAttribute",
+        "ec2:ReplaceRoute",
+        "ec2:CreateRoute",
+        "ec2:DeleteRoute",
+        "ec2:AssociateAddress",
+        "ec2:DisassociateAddress",
+
+        "s3:CreateBucket",
```
xoen: How is SoftNAS using S3?

infra/terraform/global/outputs.tf
kerin: actually no, I'll remove it

infra/terraform/global/iam.tf
kerin: it's storing a lockfile to manage primary/secondary failover

infra/terraform/modules/user_nfs_softnas/ec2.tf
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  ingress {
+    from_port = 8
+    to_port = 0
+    protocol = "icmp"
+    security_groups = ["${var.bastion_security_group_id}"]
+  }
+
+  ingress {
```
xoen: What is this for?

infra/terraform/modules/user_nfs_softnas/ec2.tf
```diff
@@ -0,0 +1,142 @@
...
+    from_port = 0
+    to_port = 0
+    protocol = "-1"
+    self = true
+  }
+
+  egress {
```
xoen: Another probably stupid question, what's this?

infra/terraform/environments/dev/main.tf
```diff
@@ -57,6 +57,19 @@ module "user_nfs" {
     performance_mode = "maxIO"
 }
 
+module "user_nfs_softnas" {
+    source = "../../modules/user_nfs_softnas"
+
+    env = "${var.env}"
```
xoen: I know the default for the variable `softnas_role_name` is `SoftNAS_HA_IAM` so shouldn't matter but I wonder if we should pass it in as `"${aws_iam_role.softnas}"`. Also, any reason to not create this role/policy inside the module (as far as I can't tell it's not used anywhere else outside it).

infra/terraform/global/iam.tf
```diff
@@ -25,3 +25,76 @@ EOF
 resource "aws_iam_access_key" "auth0_ses" {
     user = "${aws_iam_user.auth0_ses.name}"
 }
+
+resource "aws_iam_role" "softnas" {
```
xoen: I wonder if we shouldn't have separate roles for the different environments. We usually do for other modules, it may be overkill in this case.

infra/terraform/modules/user_nfs_softnas/ec2.tf
```diff
@@ -0,0 +1,142 @@
+data "aws_ami" "softnas" {
+  most_recent = true
```
xoen: I'm not very familiar with AMI, but this would basically get the last image of SoftNAS right? I wonder if we should stick with a version.

infra/terraform/modules/user_nfs_softnas/main.tf
```diff
@@ -0,0 +1,31 @@
...
+variable "num_instances" {
+    default = 2
+}
+
+variable "instance_type" {
+    default = "m4.large"
+}
+
+variable "nfs_mountpoint_ip" {
+    default = "172.16.0.1"
```
xoen: Is this always the same?

xoen:
@kerin I had a look and I didn't spot anything major, see my comments/questions. But I'm probably the wrong person to review this.

infra/terraform/global/iam.tf
kerin: Unfortunately we can't - SoftNAS looks for this policy by name, so it has to be shared across environments.

infra/terraform/global/iam.tf
xoen: 👍

infra/terraform/global/iam.tf
xoen: Interesting 👍

infra/terraform/global/iam.tf
xoen: Oh I see, thanks for the clarification 👍

infra/terraform/environments/dev/main.tf
kerin: This is really more of a constant than a variable, as SoftNAS requires that the role has this name in all environments. I tried to reference the role name from the global resource definition with `${data.terraform_remote_state.base.softnas_role_name}`, but I ran into an issue with Terraform not seeing that output variable when using the terraform `-target` flag to restrict `plan` and `apply` to just this module, so ended up hardcoding it instead.

infra/terraform/modules/user_nfs_softnas/ec2.tf
kerin: We wouldn't redeploy SoftNAS to update it, as it handles software updates itself via the GUI - the `most_recent` flag here is just to ensure that you get the most up-to-date version when you deploy it for the first time.

infra/terraform/environments/dev/main.tf
xoen: OK, it's not worth changing anything then and as I said we're already covered by the default value 👍

infra/terraform/modules/user_nfs_softnas/ec2.tf
kerin: That allows the softnas instances to communicate with each other via any protocol on any port - they communicate via SSH, FTP, NFS and ICMP to handle replication and HA failover.

infra/terraform/modules/user_nfs_softnas/ec2.tf
xoen: Nice 👍

infra/terraform/modules/user_nfs_softnas/ec2.tf
kerin: That allows the instances to send outbound traffic to any destination - without that they wouldn't be able to make outbound internet requests.

infra/terraform/modules/user_nfs_softnas/ec2.tf
xoen: Ah OK, I was wondering why we allowed ICMP, I thought it was for us (humans) but it's probably used to periodically and automatically check the other node is alive 👍

infra/terraform/modules/user_nfs_softnas/main.tf
kerin: Yes - it's an IP outside of the IP range we use elsewhere, so it won't collide with anything. This is how SoftNAS handles failover - it adds a rule to the routing table to direct traffic for that IP to the current active SoftNAS primary - when failover happens and the secondary takes over it updates the routing table to send traffic for that IP to itself.

PR #65, opened 2017-05-31
Create S3 bucket on GitHub team creation

xoen:
## What

This code adds the `team_events()` lambda function which handles SNS messages in the `${ENV}_github_team_events` topic ([GitHub team events](https://developer.github.com/v3/activity/events/types/#teamevent)).

We're interested in the creation and deletion of teams in GitHub:
* When a team is created a corresponding private S3 bucket and some policies to access it are created
* When a team is deleted the corresponding IAM policies are deleted (but not the bucket!)

## Other lambda functions involved

The `team_events` lambda is relatively dumb and invoke other lambdas to perform the required operations: 
* `create_team_bucket()` to create the S3 bucket (**In this PR**)
* `create_team_bucket_policies()` to create the IAM policies to access this bucket
* `delete_team_bucket_policies()` to delete the team bucket IAM policies

**NOTE**: The last two lambda functions will follow.

## Ticket
https://trello.com/c/UZ8qooqO/116-xl-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create-update-delete-aws-resources


infra/terraform/modules/data_access/team_events/requirements.dev.txt
```diff
@@ -0,0 +1,5 @@
+boto3
+
+pytest
+pytest-spec
+pytest-watch
```
andyhd: Do we need a requirements file for each directory of lambda functions? In dev, we could just have one virtualenv for all of them, couldn't we? Since there's only `requirements.dev.txt`s, I'm assuming there's no need for a requirements file in production, right? The lambda python environment has boto3 preloaded?

infra/terraform/modules/data_access/team_events/tests/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+        "LAMBDA_CREATE_TEAM_BUCKET_ARN": TEST_LAMBDA_CREATE_TEAM_BUCKET_ARN,
+        "LAMBDA_CREATE_TEAM_BUCKET_POLICIES_ARN": TEST_LAMBDA_CREATE_TEAM_BUCKET_POLICIES_ARN,
+        "LAMBDA_DELETE_TEAM_BUCKET_POLICIES_ARN": TEST_LAMBDA_DELETE_TEAM_BUCKET_POLICIES_ARN,
+    }):
+        yield
+
+
+@pytest.fixture
+def lambda_client_mock():
+    return mock.create_autospec(boto3.client("lambda"))
```
andyhd: This fixture is common to all our lambda functions, so can we have all our lambda functions in one python module and have code and tests shared between them?

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -0,0 +1,42 @@
...
+import json
+import logging
+import os
+
+import boto3
+
+
+LOG = logging.getLogger(__name__)
+LOG_LEVEL = os.environ.get("LOG_LEVEL", "DEBUG")
+LOG.setLevel(LOG_LEVEL)
```
andyhd: Logging code could be shared by all lambda functions, for example

andyhd:
Looks good :+1:, just some questions/comments

infra/terraform/modules/data_access/team_events/requirements.dev.txt
xoen: Correct, I'm not using any external module apart from `boto3` which is already available in AWS lambda. We could have a python virtualenv for the `data_access` module I guess.

xoen: Done in this commit: https://github.com/ministryofjustice/analytics-platform-ops/pull/65/commits/c344e57b5c3d64cf0a2610a09bb3d036523d1a89

xoen:
Thanks for your review @andyhd - I know have a single `requirements.dev.txt`/virtualenv for the whole `data_access` terraform module.

I have mixed feelings about having a single module for all the lambda function as I like the fact that these are self contained and independent. Some of these lambda functions are already (and will more) part of a single module (e.g. `create_user_role()`/`delete_user_role()` which are related and part of the `users` module). Also, have all in one module would trigger the re-deploy of all the lambda function when any of them change.

That said, I agree with you that we can probably do better. I will merge this and see if I can think about a better way. We can also pair and this discuss this later at some point.

PR #66, opened 2017-06-01
Add lambda function to create team bucket policies

xoen:
## What

This lambda function will create a read-only IAM policy and a read/write IAM policy.

The read-only policy has the following permissions:
 - list buckets in AWS console
 - list objects in the team bucket
 - read objects in the team bucket

The read/write policy has all the permissions in the read-only one plus:
 - can delete/rename/update/restore objects and their ACL in the team bucket

## When/How is called?

This lambda function is invoked asyncronously by the `team_event`
lambda when a team is created.

## How these IAM policies will be used?

When a user will later be added to this team, the right policy will be attached to his role:
 - read-only if 'member'
 - read/write if 'owner' or 'mainteiner'

## Ticket

https://trello.com/c/UZ8qooqO/116-(xl)-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create%2Fupdate%2Fdelete-aws-resources

### DIFF
--- a/infra/terraform/modules/data_access/teams/teams.py
+++ b/infra/terraform/modules/data_access/teams/teams.py
@@ -38,5 +38,96 @@ def create_team_bucket(event, context):
     )
 
 
+def create_team_bucket_policies(event, context):
+    """
+    Creates the policies for the team S3 bucket:
+      - a "readonly" policy and
+      - a "readwrite" policy
+
+    event = {"team": {"slug": "justice-league"}}
+    """
+    bucket = bucket_name(event["team"]["slug"])
+
+    create_policy(
+        name="{}-readonly".format(bucket),
+        bucket=bucket,
+        readwrite=False,
+    )
+
+    create_policy(
+        name="{}-readwrite".format(bucket),
+        bucket=bucket,
+        readwrite=True,
+    )
+
+
+def create_policy(name, bucket, readwrite):
+    LOG.debug("Creating '{}' policy".format(name))
+
+    policy_document = get_policy_document(bucket, readwrite)
+
+    client = boto3.client("iam")
+    client.create_policy(
+        PolicyName=name,
+        Path="/teams/",
+        PolicyDocument=json.dumps(policy_document),
+    )
+
+
+def get_policy_document(bucket, readwrite):
+    bucket_arn = "arn:aws:s3:::{}".format(bucket)
+
+    statements = [
+        {
+            "Sid": "ListBucketsInConsole",
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetBucketLocation",
+                "s3:ListAllMyBuckets"
+            ],
+            "Resource": "arn:aws:s3:::*"
+        },
+        {
+            "Sid": "ListObjects",
+            "Action": [
+                "s3:ListBucket"
+            ],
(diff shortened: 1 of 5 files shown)


xoen:
(rebased)

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -38,5 +38,96 @@ def create_team_bucket(event, context):
...
+    )
+
+    create_policy(
+        name="{}-readwrite".format(bucket),
+        bucket=bucket,
+        readwrite=True,
+    )
+
+
+def create_policy(name, bucket, readwrite):
```
andyhd: The `readwrite` parameter doesn't actually provide any more information than you already have in `name` - whether the policy should be read-only or read-write.

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -38,5 +38,96 @@ def create_team_bucket(event, context):
...
+    LOG.debug("Creating '{}' policy".format(name))
+
+    policy_document = get_policy_document(bucket, readwrite)
```
andyhd: I'd be tempted to inline this rather than assign to a variable - it's not a long line or difficult to understand

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -38,5 +38,96 @@ def create_team_bucket(event, context):
...
+                "s3:GetObject",
+                "s3:GetObjectAcl",
+                "s3:GetObjectVersion",
+            ],
+            "Effect": "Allow",
+            "Resource": "{}/*".format(bucket_arn)
+        },
+    ]
+
+    if readwrite:
```
andyhd: If you passed in `name` rather than `readwrite`, this test could be `if name.endswith('-readwrite')`, which is still easy to understand.

infra/terraform/modules/data_access/teams/tests/test_create_team_bucket_policies.py
```diff
@@ -0,0 +1,32 @@
+import json
+from mock import call
+
+import pytest
+
+import teams
+
+from tests.conftest import TEST_TEAM_SLUG, TEST_READONLY_POLICY_DOCUMENT, \
+    TEST_READWRITE_POLICY_DOCUMENT, TEST_BUCKET_NAME
```
andyhd: Rather than using `\` to continue a line, wrap the list of import in parentheses, and maybe put one item per line, eg:
```python
from tests.conftest import (
    TEST_TEAM_SLUG,
    ...
)
```

xoen: Nice, thanks.

infra/terraform/modules/data_access/teams/teams.py
xoen: I thought about that but I wanted to avoid the call within the `json.dumps()` call. I guess it's still readable enough so I'll inline it 👍

xoen:
@andyhd https://github.com/ministryofjustice/analytics-platform-ops/pull/66/commits/72801908eff888798c01e7f2c06054121f16598c

xoen:
Thanks for the feedback Andy!

PR #67, opened 2017-06-05
Use python create_user script installed on docker image

andyhd:
## What

Change create_user Jenkins job to use new python script which checks that user is a member of the moj-analytical-services organization

See [https://trello.com/c/jlT5g9DC](Bugfix: Ensure user is part of analytical services org as part of deployment)

### DIFF
--- a/jenkinsfiles/create_user.groovy
+++ b/jenkinsfiles/create_user.groovy
@@ -13,6 +13,6 @@ node {
     }
 
     stage ("Create platform user") {
-        sh "jenkinsfiles/create_user.sh ${env.PLATFORM_ENV} ${env.USERNAME} ${env.EMAIL} \"${env.FULLNAME}\""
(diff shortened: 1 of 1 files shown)


jenkinsfiles/create_user.groovy
```diff
@@ -13,6 +13,6 @@ node {
     }
 
     stage ("Create platform user") {
-        sh "jenkinsfiles/create_user.sh ${env.PLATFORM_ENV} ${env.USERNAME} ${env.EMAIL} \"${env.FULLNAME}\""
+        sh "/usr/local/bin/create_user ${env.USERNAME} ${env.EMAIL} --env ${env.PLATFORM_ENV} --fullname \"${env.FULLNAME}\""
```
xoen: Looks good to me. I had a question about the use of `--` for all these arguments in the other PR: https://github.com/ministryofjustice/analytics-platform-jenkins-agent/pull/4/files#r120372836

xoen:
Looks good to me, I wonder if we should just use `--` for all arguments instead of mix and match, but there may be a reason why you wanted to make this distinction 👍

PR #68, opened 2017-06-06
Add lambda function to delete team bucket policies

xoen:
## What

When a team is deleted in GitHub we want the corresponding S3 bucket
policies to be deleted, but not the bucket itself which may still
contain important data.

Delete a policy is not as straightforward as it sounds unfortunately
because it needs to be detached from roles, groups and users first
([See AWS documentation](https://docs.aws.amazon.com/de_de/IAM/latest/APIReference/API_DeletePolicy.html))

## Ticket

https://trello.com/c/UZ8qooqO/116-(xl)-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create%2Fupdate%2Fdelete-aws-resources

### DIFF
--- a/infra/terraform/modules/data_access/teams/teams.py
+++ b/infra/terraform/modules/data_access/teams/teams.py
@@ -1,6 +1,7 @@
 '''
 Environment variables:
  - BUCKET_REGION, region where bucket will be created, e.g. "eu-west-1"
+ - IAM_ARN_BASE, e.g. "arn:aws:iam::1234"
  - STAGE, e.g. "dev", "alpha", etc...
 '''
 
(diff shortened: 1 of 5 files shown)


infra/terraform/modules/data_access/teams/tests/test_delete_team_bucket_policies.py
```diff
@@ -0,0 +1,77 @@
...
+    TEST_TEAM_SLUG,
+    TEST_USER_NAME
+)
+
+
+@pytest.mark.usefixtures(
+    "given_the_env_is_set",
+    "given_iam_is_available",
+)
+def test_when_the_team_is_deleted_the_bucket_policies_are_detached_from_anything(iam_client_mock):
```
andyhd: `s/anything/everything/`

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -126,5 +127,91 @@ def get_policy_document(bucket_name, readwrite):
...
+
+def detach_policy_from_user(policy_arn, user_name):
+    LOG.debug("Detaching policy '{}' from user '{}'".format(
+        policy_arn, user_name
+    ))
+    client = boto3.client("iam")
+    client.detach_user_policy(
+        UserName=user_name,
+        PolicyArn=policy_arn,
+    )
```
andyhd: You could also remove the repetition of similar functions here by rolling everything back into `detach_policy_from_entities`, as the only things that really change are the name of the boto client method and the name of the entity name parameter. Eg:
```python
entity_types = {
    'PolicyRoles': {'method': 'detach_role_policy', 'name_field': 'RoleName'},
    ...
}

for entity_type in entity_types:
    name_field = entity_type['name_field']
    for entity in entities[entity_type]:
        detach = client.getattr(entity_type['method'])
        detach(PolicyArn=policy_arn, **{name_field: entity[name_field]})
```
Although this is more difficult to read, so again, I'm not sure it's better.

infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -126,5 +127,91 @@ def get_policy_document(bucket_name, readwrite):
...
+    LOG.debug("Policy is attached to: {}".format(json.dumps(entities)))
+
+    for role in entities["PolicyRoles"]:
+        detach_policy_from_role(policy_arn, role["RoleName"])
+
+    for group in entities["PolicyGroups"]:
+        detach_policy_from_group(policy_arn, group["GroupName"])
+
+    for user in entities["PolicyUsers"]:
+        detach_policy_from_user(policy_arn, user["UserName"])
```
andyhd: You could reduce the repetition here by having a lookup, eg:
```python
entity_types = {
    'PolicyRoles': {'detach_fn': detach_policy_from_role, 'name_field': 'RoleName'},
    ...
}

for entity_type in entity_types:
    for entity in entities[entity_type]:
        entity_type['detach_fn'](policy_arn, entity[entity_type['name_field']])
```
but I'm not sure that would be better.

xoen:
Thanks for your review and the interesting tricks @andyhd - I will amend the commit with the fix for that typo/grammar error but I will leave the rest as it is for simplicity/readability as discussed.

PR #69, opened 2017-06-06
Adds AWS Auth0 identity provider

kerin:
## What

Adds a small Terraform module to create an IAM identity provider using the SAML metadata provided by Auth0.

This works in conjunction with [this Auth0 rule](https://github.com/ministryofjustice/analytics-platform-auth0/blob/dev/rules/aws-saml-role-mapping.js) which maps Auth0/IDP attributes to SAML attributes.


.gitattributes
```diff
@@ -7,3 +7,4 @@ k8s-resources/jenkins/envs/**/* filter=git-crypt diff=git-crypt
 k8s-resources/prometheus/envs/**/* filter=git-crypt diff=git-crypt
 infra/kops/clusters/**/* filter=git-crypt diff=git-crypt
 chart-env-config/**/* filter=git-crypt diff=git-crypt
+infra/terraform/modules/federated_identity/saml/**/* filter=git-crypt diff=git-crypt
```
xoen: What is this for?

xoen: Oh is it because the `auth0-metadata.xml` file has sensitive information in it?

kerin: SAML metadata includes an x509 certificate, so this file needs to be encrypted

xoen: 👍

xoen:
A terraform style note, I started to use the convention of keeping module input variables in a `input.tf` file and module output variables in a `output.tf` file. I kind of like the idea of having well defined places where to look at to understand how to interact with a module but maybe I'm talking rubbish.

Anyway, looks good to me 👍  - I'll update the `data_access` module to use this module's `saml_provider_arn` output variable.

xoen:
Seal of approval

PR #70, opened 2017-06-06
SAML integration

xoen:
## What

Updated code to:
1. `data_access` module use `saml_provider_arn` output variable from `federated_identity` module instead of hardcoding the SAML provider ARN
2. Removed `_role` suffix from generated roles to adhere with [Auth0 rule](https://github.com/ministryofjustice/analytics-platform-auth0/blob/dev/rules/aws-saml-role-mapping.js#L3-L17) used to generate role assumed on login.

## Related PR

[Adds AWS Auth0 identity provider](https://github.com/ministryofjustice/analytics-platform-ops/pull/69)

### DIFF
--- a/infra/terraform/environments/dev/main.tf
+++ b/infra/terraform/environments/dev/main.tf
@@ -126,6 +126,8 @@ module "data_access" {
     env = "${var.env}"
     account_id = "${data.aws_caller_identity.current.account_id}"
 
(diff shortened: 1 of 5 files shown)


xoen:
@kerin FYI, it's pretty much what discussed earlier via chat.

PR #71, opened 2017-06-08
Use per-env Auth0 credentials

andyhd:
## What

Use Auth0 credentials per environment (dev, alpha) - standardised credentials names across jenkins CI and control panel jenkins.


xoen:
Nice 👍

PR #72, opened 2017-06-08
Add lambda function to handle GitHub membership events

xoen:
Handles events coming from GitHub/membership SNS topic.

These are published when a user is added/removed to/from a team in
GitHub:
 - If a user is added to a group, we attach the corresponding bucket policy
   to this user's role
 - If a user is removed to a group, we detach the corresponding bucket policy
   from this user's role

These two operations are actually performed by two independent lambda
functions injected via the `LAMBDA_ATTACH_BUCKET_POLICY_ARN` and
`LAMBDA_DETACH_BUCKET_POLICY_ARN` environment variables.

**NOTE**: These lambda functions currently don't exist and I'll update the
terraform resources later to pass the actual ARNs. This shouldn't change
the logic.

## Limitations

Surprising GitHub webhook membership events payload don't include the user
role for the team where is added. And also, no events are published when
a user's role changes.

This means we don't have a way to tell if a user is just a "member",
"maintainer" or "owner" in a team or when its level changes.

For simplicity we'll add the read/write policy to the user.
When/if we'll have a UI to handle permissions we can add extra logic to
attach a readonly policy when more appropriare.

## Resources

* [Ticket](https://trello.com/c/UZ8qooqO/116-xl-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create-update-delete-aws-resources)
* ["membership" webhook event GitHub documentation](https://developer.github.com/v3/activity/events/types/#membershipevent)

### DIFF
--- a/infra/terraform/environments/dev/main.tf
+++ b/infra/terraform/environments/dev/main.tf
@@ -128,6 +128,7 @@ module "data_access" {
 
     saml_provider_arn = "${module.federated_identity.saml_provider_arn}"
 
+    membership_events_topic_arn = "${module.notifications.membership_events_topic_arn}"
     organization_events_topic_arn = "${module.notifications.organization_events_topic_arn}"
     team_events_topic_arn = "${module.notifications.team_events_topic_arn}"
 }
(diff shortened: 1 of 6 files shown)


xoen:
Thanks for your review @andyhd

PR #74, opened 2017-06-12
Add lambda functions to attach/detach IAM policies to/from IAM roles

xoen:
### `attach_bucket_policy()`

Invoked when a member is added to a team.
Attaches the specified IAM team bucket policy to the corresponding user role (e.g. `readonly` or `readwrite`).

### `detach_bucket_policies()`

Detaches all the IAM team bucket policies (`readonly` and `readwrite`) from the user role.

One subtle thing to notice here is that a given role would probably have only one of the currently available IAM policies (e.g. `readwrite` only). As AWS raises an exception when trying to detach a policy which is not attached, we need to be sure that we handle it and continue to detach all the IAM team bucket policies from the role.

### Ticket:

https://trello.com/c/UZ8qooqO/116-xl-create-lambda-functions-to-consume-github-events-from-sns-queues-and-create-update-delete-aws-resources

### DIFF
--- a/infra/terraform/modules/data_access/membership_events/membership_events.py
+++ b/infra/terraform/modules/data_access/membership_events/membership_events.py
@@ -6,7 +6,8 @@ Environment variables:
    IAM policy for the bucket to the IAM role
  - LOG_LEVEL, change the logging level (default is "DEBUG"). Must be one of
    the python logging supported levels: "CRITICAL", "ERROR", "WARNING",
-   "INFO" or "DEBUG" (See: https://docs.python.org/2/library/logging.html#logging-levels)
+   "INFO" or "DEBUG"
+   (See: https://docs.python.org/2/library/logging.html#logging-levels)
 '''
 
 import json
@@ -26,6 +27,11 @@ LOG.setLevel(LOG_LEVEL)
 
 # "membership" webhook event: https://developer.github.com/v3/activity/events/types/#membershipevent
 def event_received(sns_event, context):
+    EVENT_HANDLERS = {
+        "added": os.environ["LAMBDA_ATTACH_BUCKET_POLICY_ARN"],
+        "removed": os.environ["LAMBDA_DETACH_BUCKET_POLICIES_ARN"],
+    }
+
     LOG.debug("SNS event received = {}".format(json.dumps(sns_event)))
     for record in sns_event["Records"]:
         event = json.loads(record["Sns"]["Message"])
@@ -34,15 +40,10 @@ def event_received(sns_event, context):
         if event["scope"] != "team":
             continue
 
-        if action == "added":
+        if action in EVENT_HANDLERS:
             invoke_lambda(
-                function=os.environ["LAMBDA_ATTACH_BUCKET_POLICY_ARN"],
-                payload=attach_payload(event),
-            )
-        elif action == "removed":
-            invoke_lambda(
-                function=os.environ["LAMBDA_DETACH_BUCKET_POLICIES_ARN"],
-                payload=detach_payload(event),
+                function=EVENT_HANDLERS[action],
+                payload=payload(event),
             )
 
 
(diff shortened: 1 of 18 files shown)


infra/terraform/modules/data_access/membership_events/membership_events.py
```diff
@@ -39,10 +34,15 @@ def event_received(sns_event, context):
...
             invoke_lambda(
-                function=EVENT_HANDLERS[action],
-                payload=payload(event),
+                function=os.environ["LAMBDA_ATTACH_BUCKET_POLICY_ARN"],
+                payload=attach_payload(event),
+            )
+        elif action == "removed":
+            invoke_lambda(
+                function=os.environ["LAMBDA_DETACH_BUCKET_POLICIES_ARN"],
+                payload=detach_payload(event),
```
andyhd: I think, rather than reintroducing the repetition that was factored out into `EVENT_HANDLERS`, you can just add the `policy` item to the payload `if action == "added"`

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -0,0 +1,100 @@
...
+    client = boto3.client("iam")
+    for policy_type in [POLICY_READ_WRITE, POLICY_READ_ONLY]:
+        # Be sure we detach all policies without stopping early
+        try:
+            client.detach_role_policy(
+                RoleName=name,
+                PolicyArn=policy_arn(team_slug, policy_type),
+            )
+        except botocore.exceptions.ClientError:
+            pass
```
andyhd: I get that we want to try to remove both types of policy, regardless of errors - but don't we care about those errors if they happen? Should they be stored instead of discarded and then raised after all removals have been tried?

infra/terraform/modules/data_access/memberships/tests/test_detach_bucket_policies.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+
+@pytest.mark.usefixtures(
+    "given_the_env_is_set",
+    "given_iam_is_available",
+)
+def test_when_invoked_detaches_all_bucket_policies_from_the_role(iam_client_mock):
+    # This is to test the scenario when 1+ of the policies tried to be
+    # detached is not actually attached: In this case we still boto/AWS will
```
andyhd: "we still boto/AWS" ?

infra/terraform/modules/data_access/memberships/tests/test_detach_bucket_policies.py
```diff
@@ -0,0 +1,58 @@
...
+    # raise an exception but we still want to carry on and detach the rest
+    # of the policies.
```
andyhd: Make this a doc-comment, rather than an inline comment

infra/terraform/modules/data_access/membership_events/membership_events.py
xoen: I'm looking at the code to see how to remove this duplication. I may come to your desk to discuss this face to face.

infra/terraform/modules/data_access/memberships/memberships.py
xoen: Well in this specific case the error is when we try to detach one of the two policies but the policy is not attached (e.g. role has only readonly policy and we try to detach readwrite).

As the alternative would be to check if the policy is attached then remove it but there is no harm in trying to detach something which is not attached. We save some logic and an additional request all the times. We may log it but it may sound alarming/cause noise without being actually useful.

infra/terraform/modules/data_access/memberships/tests/test_detach_bucket_policies.py
xoen: I need to re-read this comment again as I clearly got distracted while writing it and the sentence doesn't make any sense :P

infra/terraform/modules/data_access/memberships/tests/test_detach_bucket_policies.py
xoen: 👍

xoen:
FYI: As discussed with @kerin we now always user lowercase IAM roles/S3 bucket names: https://github.com/ministryofjustice/analytics-platform-ops/pull/74/commits/5903184daa5a9864f2a671ee8aece6d15247ba61

xoen:
@andyhd FYI: I'm making these tweaks in different branches/PRs for the sake of keeping things small.

Let me know when/if I can merge this so I can avoid to have too many things in flight at the same time ;)

PR #76, opened 2017-06-15
Adds IAM role support to Rstudio

kerin:
## What

Allows RStudio to use IAM Role-based credentials rather than explicit credential pairs.

## How to review

1. Check out this PR
2. Install or upgrade your RStudio instance to specify an IAM Role:
```
helm upgrade kerin-rstudio charts/rstudio -f chart-env-config/dev/rstudio.yml --set Username=$USERNAME --set AWS.IAMRole=users/$ENV_$USERNAME --set RStudio.Image.Tag=v1.2.3 --namespace user-$USERNAME --install --wait
```
3. Start a new R session
4. Verify that the modified aws.signature/aws.s3 libraries use the specified role


charts/kube2iam/_kube2iam/daemonset.yml
```diff
@@ -0,0 +1,32 @@
...
+    metadata:
+      labels:
+        name: kube2iam
+    spec:
+      hostNetwork: true
+      containers:
+        - image: jtblin/kube2iam:latest
+          name: kube2iam
+          args:
+            - "--base-role-arn=arn:aws:iam::593291632749:role/"
```
xoen: So I understand well this will be concatenated by `kube2iam` with the pod's annotation `iam.amazonaws.com/role` which will contain the role name.

So we probably change this to `...role/users"` unless we want to pass `users/dev_kerin`, etc.. all the times which is error prone.

xoen:
I don't know kube2iam but as far as I can see and from what I quickly read in their README this looks good.

See my comment on the roles' path but otherwise this looks good to me.

charts/kube2iam/_kube2iam/daemonset.yml
kerin: Initially I thought that too, but I think we might well end up deploying other services that use roles that won't be user-based, so it seems best to allow any role to be specified regardless of path.

kerin: ...but yes, the account ID should be a variable - I'll change that

kerin: I've just realised that the files in this directory, `_kube2iam`, are temporary files I used while developing the chart, and should be removed - the actual template files used are in `templates`

PR #78, opened 2017-06-16
Remove path from IAM roles and policies

xoen:
### Why

They don't add much value at the moment and the R library we currently use
(Cloudyr) is ignoring them causing problems.

### IAM changes

The IAM role names will now be in the format `$ENV_user_$USERNAME`.
The IAM policies will also lose the path but their name doesn't change.

### Ticket

https://trello.com/c/8hp1cSPd/229-s-get-right-of-users-path-in-dataaccess-iam-roles


xoen:
Another small one for you @andyhd  :)

This is just to remove the path from IAM roles and policies as discussed yesterday, it should help with the R part of our work.

PR #80, opened 2017-06-19
'data_access' TF module: Safer bucket names

xoen:
Team slugs coming from GitHub may contains characters which are not valid
in an S3 bucket name, e.g. underscores (`_`).

We're keeping it simple for now and just replacing the invalid characters with an hyphen (`-`).

**NOTE**: This changes doesn't handle all the S3 limitations but it should be good
enough for the time being.

See: http://docs.aws.amazon.com/en_gb/AmazonS3/latest/dev/BucketRestrictions.html

#### Ticket

https://trello.com/c/fp3eVZyG/231-s-createteambucket-lambda-to-sanitize-team-name-to-not-fail-when-containing-s


infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -213,4 +214,19 @@ def detach_policy_from_user(policy_arn, user_name):
 
 
 def bucket_name(slug):
```
andyhd: Rather than repeating this function in each lambda function subdirectory, we should probably create a python package with all our common code that can be imported.

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -114,7 +115,19 @@ def policy_arn(team_slug, policy_type):
...
+    NOTE: This is a very simple implementation which doesn't cover all the
+          S3 limitations (e.g. max length or labels limitations, etc...)
+
+    See: http://docs.aws.amazon.com/en_gb/AmazonS3/latest/dev/BucketRestrictions.html
+    '''
+
+    INVALID_BUCKET_CHARS = r"[^a-z0-9.-]"
+
+    name = team_slug.lower()
+    name = re.sub(INVALID_BUCKET_CHARS, "-", name)
```
andyhd: This code should cover at least the following S3 bucket naming rules:

* Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.).
* Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
* Bucket names must not be formatted as an IP address (e.g., 192.168.5.4).

andyhd: This code allows bucket names like `...`, `---foo---` and `1.2.3.4`

xoen: You are right. As said in the function comment, this is a simple implementation which doesn't cover all S3 limitations but should cover all the likely team names without having to store the mapping between GitHub team names and the generated S3 bucket name.

How would the code be automatically be able to transform a team called `1.2.3.4` or `...` into a valid S3 bucket?

andyhd: Sorry, yes, you are right - as discussed, the leading and trailing hyphens can be dealt with automatically, but the others cannot.

xoen:
Have another look @andyhd

andyhd:
Including Raven and all its supporting packages has added ~120 files to the repo. Is there some way we could zip them up into a single file at least? We're not interested in seeing diffs of files inside Raven and its dependencies, so just seeing opaque updates to a big binary file would be fine I think.

xoen:
I now ignore these packages so they're not added to Git as they're installed anyway by the script 👍

infra/terraform/modules/data_access/teams/teams.py
xoen: I'll address the duplication separately: https://trello.com/c/ZbjN4qxy/234-s-factor-out-lambdas-duplicated-code

xoen:
FYI: I forgot to set the `SENTRY_DSN` environment variable for the policies lambda functions. Doing it now.

xoen:
Done: https://github.com/ministryofjustice/analytics-platform-ops/pull/80/commits/becc7ca15a966e3d76e9f4f3a881f92b0062be8d

xoen:
Cheers @andyhd

PR #81, opened 2017-06-19
Set IAM role when deploying RStudio instance

xoen:
The helm chart was already using the IAM role to set the `iam.amazonaws.com/role` annotation (used by kube2iam).

The jenkins job now pass the IAM role value.
This is built from the environment and username: `$ENV_user_$USERNAME`

**SEE**: [lambda function generating the IAM roles in data_access module](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/data_access/users/users.py#L87)

#### Ticket

https://trello.com/c/QsRYlTxQ/120-s-convert-rstudio-deployments-to-use-iam-roles-instead-of-access-keys-secrets


xoen:
As discussed offline, we'll need to be careful with this as few users are using RStudio with their personal AWS credentials. This is mainly to access the CREST S3 Bucket.

We can plan a migration of the CREST data to a new S3 bucket using the new data access model and add these users to this team.

xoen:
Cheers Kerin!

PR #82, opened 2017-06-21
Add S3 logging setup to team bucket creation lambda function

kerin:
## What

Each bucket created by the team bucket lambda function requires access logs to be enabled. These changes ensure that all buckets created from github events have logging configured to send logs to a per-environment bucket, with an `s3/team-bucket-name/` prefix, to keep S3 logs separate from future CloudTrail logs, and to keep bucket logs separate from each other.

## How to review

1. Terraform apply
2. Create a team bucket via github or the Lambda console
3. Check that logging is enabled via the S3 console, or:
4. Upload a file to the team bucket and verify that logs are created (may take up to 1 hour)

### DIFF
--- a/infra/terraform/modules/data_access/teams/teams.py
+++ b/infra/terraform/modules/data_access/teams/teams.py
@@ -43,7 +43,7 @@ def send_exceptions_to_sentry(fn):
 @send_exceptions_to_sentry
 def create_team_bucket(event, context):
     """
-    Creates the team's S3 bucket
+    Creates the team's S3 bucket, with logging enabled
 
     event = {"team": {"slug": "justice-league"}}
     """
@@ -61,6 +61,15 @@ def create_team_bucket(event, context):
         Bucket=name,
         ACL="private",
         CreateBucketConfiguration={"LocationConstraint": region},
+    )
+    client.put_bucket_logging(
+        Bucket=name,
+        BucketLoggingStatus={
+            'LoggingEnabled': {
+                'TargetBucket': os.environ.get('LOGS_BUCKET_NAME'),
+                'TargetPrefix': "s3/{}/".format(name)
+            }
+        },
     )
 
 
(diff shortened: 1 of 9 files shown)


infra/terraform/modules/data_access/teams/teams.py
```diff
@@ -62,6 +62,15 @@ def create_team_bucket(event, context):
         ACL="private",
         CreateBucketConfiguration={"LocationConstraint": region},
     )
+    client.put_bucket_logging(
+        Bucket=name,
+        BucketLoggingStatus={
+            'LoggingEnabled': {
+                'TargetBucket': os.environ.get('LOGS_BUCKET_NAME'),
```
andyhd: What happens if `LOGS_BUCKET_NAME` is `None` (the default return value of `get` if the env var is not set? Should this be `os.environ['LOGS_BUCKET_NAME']` instead, to raise a `KeyError` exception if it is not set?

infra/terraform/modules/data_access/teams/tests/test_create_team_bucket.py
```diff
@@ -1,10 +1,10 @@
-import json
-
 import pytest
 
 import teams
 
-from tests.conftest import TEST_TEAM_SLUG, TEST_BUCKET_NAME, TEST_BUCKET_REGION
+from tests.conftest import (TEST_TEAM_SLUG, TEST_BUCKET_NAME,
+                            TEST_BUCKET_REGION, TEST_LOGS_BUCKET_NAME,
+                            TEST_LOGS_PREFIX)
```
andyhd: If you're going to break a list into multiple lines, it's better to have one item per line

andyhd: And not indent to an arbitrary column - break after the open bracket and indent one level

andyhd:
Are you going to squash all these commits?

kerin:
@andyhd no

PR #83, opened 2017-06-21
Factor out common Lambdas code (sentry/naming) into modules

xoen:
The lambda functions in the `data_access` terraform module had some
duplication, specifically:
 - bucket_name
 - policy_name/arn
 - role_name
 - sending of exceptions to sentry

I extracted this code into 2 modules: `naming` and `sentry`.

These are local modules and they need to be installed into the lambda
directories before the archive is created, this is done by using a terraform's
`null_resource`/`local-exec` which executes a `build.sh` script.

This `build.sh` script install all required modules in the lambda root
directory.

**SEE**: http://docs.aws.amazon.com/de_de/lambda/latest/dg/lambda-python-how-to-create-deployment-package.html

### Also renamed environment variable `STAGE` => `ENV`

We use the word "env" across the board so "stage" can be consfusing.
Changing it for consistency sake.

### Ticket

https://trello.com/c/ZbjN4qxy/234-s-factor-out-lambdas-duplicated-code


infra/terraform/modules/data_access/common/sentry/sentry/sentry.py
```diff
@@ -0,0 +1,26 @@
...
+        except Exception:
+            client = Sentry(
+                os.environ["SENTRY_DSN"],
+                transport=SentryHTTPTransport,
+                environment=os.environ["ENV"],
+            )
+            client.captureException()
+            raise
+
+    return wrapped
```
andyhd: This path `common/sentry/sentry/sentry.py` seems like a code smell to me. Is there any reason we can't have the `sentry` module and the `naming` module in the same package with a single `setup.py`? We could call it `moj_analytics_tools` (or something) instead of `common` and then `from moj_analytics_tools import sentry` or `from moj_analytics_tools.sentry import catch_exceptions`. I don't think the size of the combined modules would be too big to upload with the lambda functions.

infra/terraform/modules/data_access/common/sentry/sentry/sentry.py
```diff
@@ -0,0 +1,26 @@
...
+ - SENTRY_DSN, Sentry DSN
+'''
+
+import os
+
+from raven import Client as Sentry
+from raven.transport.http import HTTPTransport as SentryHTTPTransport
+
+
+def catch_exceptions(fn):
```
andyhd: Minor naming quibble - to "catch" an exception suggests that it is handled, but this re-raises. Maybe `report_exceptions` or `log_exceptions` would be more self-explanatory.

andyhd:
I don't think the Python code should live here, buried in the terraform files. We should move it to a separate repository. You could use the `build.sh` script to fetch a zip file from the repository and upload it to AWS.

infra/terraform/modules/data_access/common/sentry/sentry/sentry.py
xoen: I'll go for `report_exceptions`, thanks 👍

infra/terraform/modules/data_access/common/sentry/sentry/sentry.py
xoen: I've seen this before and I get why it makes sense in the beginning. But then this soon become a dumping ground for "anything common". I like to keep packages small with the potential for having them potentially reused independently one day.

xoen: I'll move the code in the `__init__.py`s, although one benefit of it is that you can explicitly decide what to expose to the client code and what not. But for now probably these are simple enough that code can happily stay all in the `__init__.py`s files.

xoen:
True, we can improve this, we may decide to move these modules in independent repositories and change the requirements/pip commands to install from Git instead of from filesystem.

xoen:
@andyhd I moved the common modules' code in the `__init__.py`: https://github.com/ministryofjustice/analytics-platform-ops/pull/83/commits/a758745455ad468c62a6e1ed760f9ff62f3175e0

infra/terraform/modules/data_access/common/sentry/sentry/sentry.py
andyhd: you can control what is exposed using the `__all__` list - this is more effective and explicit than importing from a separate file, because you could still import everything directly from that file. eg:
```python
__all__ = ['catch_exceptions']

def catch_exceptions(fn):
    ...

def something_that_wont_be_exposed():
    ....
```

xoen: Good to know 👍

xoen:
Cheers @andyhd

PR #84, opened 2017-06-26
Add Shiny apps support for no authentication and/or IP-based restrictions

xoen:
## What

This is to use the new version of the Auth Proxy which adds support for:
* Optionally disabling authentication, e.g. user will not have to login, when `AUTHENTICATION_REQUIRED = false`, by default authentication is enabled
* Optionally check user's IP (`IP_RANGES` environment variable)

You can mix and match so you can have all combinations of the two, examples:
 * No authentication and IP is not checked (public shiny app)
 * No authentication but IP is checked (shiny app shared with everyone in a building)
 * Authentication and no IP is checked (as before, user needs to login to see the shiny app and can see it from anywhere, **this is the default**)
 * Authentication and IP is checked (user needs to login AND can only see shiny app from certain locations)

## Related Resources

* [Proxy code changes](https://github.com/ministryofjustice/analytics-platform-auth-proxy/pull/2/files)
* [Docker image @ quay.io](https://quay.io/repository/mojanalytics/auth-proxy?tab=tags)

## Ticket

https://trello.com/c/UVgHMv1C/223-l-allow-ip-based-access-control-to-shiny-apps-for-dom1-quantum-etc-instead-of-or-as-well-as-user-based-access-controls


xoen:
@kerin FYI: I changed this as the [proxy now gets IP ranges using CIDR blocks](https://github.com/ministryofjustice/analytics-platform-auth-proxy/pull/2/files) instead of a regular expression.

xoen:
Cheers Kerin

PR #85, opened 2017-06-28
Shiny helm chart: Fix for "upgrade failed expected string got bool"

xoen:
I'm see this error in Jenkins and it's likely caused by the fact that
the `AuthProxy.AuthenticationRequired` value is `true` (boolean) instead of
`"true"` (string)

Error:

    Error: UPGRADE FAILED: error validating "": error validating data: expected type string, for field spec.template.spec.containers[0].env[7].value, got bool

## Ticket

Part of https://trello.com/c/UVgHMv1C/223-l-allow-ip-based-access-control-to-shiny-apps-for-dom1-quantum-etc-instead-of-or-as-well-as-user-based-access-controls


xoen:
Shown to Kerin offline 👍

PR #88, opened 2017-07-12
Ship RStudio access and error logs to external elasticsearch cluster

kerin:
## What

As part of moving to an external, standard Elasticsearch cluster in place of AWS's Elasticsearch service, additional config options are required for fluentd, added as part of this PR.

Additionally:
  - RStudio log tags have been updated to include the user's username
  - Kubernetes healthcheck requests have been excluded, to minimise noise in our logging server

## How to review

Deploy/update your RStudio instance, then log into Kibana to check that access logs appear, tagged with your username.


charts/rstudio/templates/fluentd-configmap.yml
```diff
@@ -8,33 +8,39 @@ metadata:
...
-      read_from_head true
-      tag rstudio-error
+      format none
+      tag "#{ENV['USER']}.rstudio.error"
     </source>
 
+    # Ignore k8s healthcheck requests
+    <filter *.nginx.access>
+      @type grep
+      exclude1 agent ^Go-http-client\/.*$
```
xoen: As discussed offline, from what I can tell in this case the healtcheck endpoint is `/health-check`, so we may prefer to filter by `path` field:

```ruby
exclude1 path ^\/health-check$
```

charts/rstudio/values.yaml
```diff
@@ -9,13 +9,16 @@ AWS:
...
+  Scheme: "http"
+  Username: ""
+  Password: ""
 Fluentd:
   Image:
-    Repository: quay.io/fluent/fluentd-kubernetes-daemonset
-    Tag: latest
-    PullPolicy: "IfNotPresent"
+    Repository: fluent/fluentd-kubernetes-daemonset
+    Tag: elasticsearch
```
xoen: In my case I used the `stable-elasticsearch` tag which at this time should be equivalent to `elasticsearch` but I thought was less likely to break.

scripts/upgrade_rstudio_pods.sh
```diff
@@ -1,27 +1,17 @@
 #!/usr/bin/env bash
 set -ex
 
-if [ $# -lt 1 ]; then
-  echo 1>&2 "$0: Arguments: DOCKER_TAG"
-  exit 2
-fi
-
-DOCKER_TAG=$1
+DOCKER_TAG=${1:-"latest"}
```
xoen: What is [this sorcery](https://stackoverflow.com/questions/27445455/what-does-the-colon-dash-mean-in-bash) - nice 👍

scripts/upgrade_rstudio_pods.sh
```diff
@@ -1,27 +1,17 @@
...
-
-    access_key=$(kubectl get secret $secret_name -n $namespace -o jsonpath="{.data.aws_access_key_id}" | base64 --decode)
-    secret_key=$(kubectl get secret $secret_name -n $namespace -o jsonpath="{.data.aws_secret_access_key}" | base64 --decode)
 
     helm upgrade $user-rstudio charts/rstudio \
         -f chart-env-config/$env/rstudio.yml \
         --set Username=$user \
-        --set AWS.AccessKeyId=$access_key \
-        --set AWS.SecretAccessKey=$secret_key \
+        --set AWS.IAMRole=${env}_user_${user} \
```
xoen: Odd, I thought we got rid of these few weeks (months?) ago.

xoen:
Looks good to me, feel free to filter by `/health-check` `path` if you want.

scripts/upgrade_rstudio_pods.sh
kerin: we did in the Jenkins jobs etc, but this is a little utility script to forcibly redeploy all instances, and clearly I never actually checked this change in

xoen: Uh, sorry, I didn't realise it was that upgrade script. I thought was the script to deploy rstudio instances 👍

PR #89, opened 2017-07-13
attach_bucket_policy lambda function retries twice before notifying Sentry

xoen:
## What (NOW)

We now catch this specific exception and retry twice after a brief amount of
time (200ms, tested). This prevent the lambda failing and the error to get
to us unless can't be handled this way.

## Why (BEFORE)

Because of the asyncronous nature of these lambda functions, when creating
a new team or the user is new to the organisation:
1) The team bucket's IAM policy may not be created yet
2) The user's IAM role may not be created yet

In these two cases the `attach_bucket_policy` lambda function would raise
an [`NoSuchEntity` exception](https://sentry.service.dsd.io/mojds/data-access-lambda/issues/5251/events/146927/).
AWS will automatically retry and succeed but we'll be notified about this
error via Sentry.

## Benefits

Add a retry/sleep is not the pretties thing but:
1) From my manual experiments I've noticed that sleeping and retrying actually
   improve the overall execution time. This is because not retrying will
   cause a HTTP request to Sentry which increases the total execution time
   (sometimes ~1000-1100ms), plus the execution time on the AWS retry (~700ms).
   Conversely, sleep 200ms would cause the lambda function to succeed and
   avoid the HTTP call to Sentry and the AWS retry bringing the total execution
   time when this exception happen to ~1300ms (from 1700ms = 1000ms + 700ms)
2) We have a very low number of invocation and we don't exceed the free tier
3) Have false positive sentry errors lead to ignore the actual errors

## Ticket

https://trello.com/c/3VbPHbM9/273-retry-mechanism-for-dataaccess-lambdas

### DIFF
--- a/infra/terraform/modules/data_access/lambda_memberships.tf
+++ b/infra/terraform/modules/data_access/lambda_memberships.tf
@@ -5,8 +5,7 @@ resource "null_resource" "memberships_install_deps" {
     }
 
     triggers {
-        build_sh_sha = "${sha256(file("${path.module}/memberships/build.sh"))}"
-        requirements_sha = "${sha256(file("${path.module}/memberships/requirements.txt"))}"
+        force_rebuild = "${timestamp()}"
     }
 }
 
--- a/infra/terraform/modules/data_access/memberships/.gitignore
+++ b/infra/terraform/modules/data_access/memberships/.gitignore
@@ -1,0 +1,4 @@
+contextlib2*
+naming*
+raven*
+sentry*
--- a/infra/terraform/modules/data_access/memberships/memberships.py
+++ b/infra/terraform/modules/data_access/memberships/memberships.py
@@ -63,7 +63,7 @@ def attach_bucket_policy(event, context):
     retry(fn)
 
 
-def retry(fn, max_attempts=5, delay=0.150):
+def retry(fn, max_attempts=10, delay=0.200):
     attempts = 0
     while True:
         try:
--- a/infra/terraform/modules/data_access/memberships/memberships.py
+++ b/infra/terraform/modules/data_access/memberships/memberships.py
@@ -54,15 +54,20 @@ def attach_bucket_policy(event, context):
 
     client = boto3.client("iam")
 
-    # If role or policy are not there yet, retry
-    RETRY_DELAY = 0.200
-    tries = 2
+    def fn():
+        client.attach_role_policy(
+            RoleName=naming.role_name(username),
+            PolicyArn=policy_arn(team_slug, policy_type),
+        )
+
+    retry(fn)
+
+
+def retry(fn, attempts=5, delay=0.150):
+    tries = attempts
     while tries > 0:
         try:
-            client.attach_role_policy(
-                RoleName=naming.role_name(username),
-                PolicyArn=policy_arn(team_slug, policy_type),
-            )
+            fn()
             break
         except botocore.exceptions.ClientError as error:
             if error.response["Error"]["Code"] != "NoSuchEntity":
@@ -73,8 +78,8 @@ def attach_bucket_policy(event, context):
             LOG.warning(
                 "error while attaching role to policy: {}.".format(error))
             if tries > 0:
-                LOG.warning("Retry in {}s...".format(RETRY_DELAY))
-                time.sleep(RETRY_DELAY)
+                LOG.warning("Retry in {}s...".format(delay))
+                time.sleep(delay)
             else:
                 raise
 
--- a/infra/terraform/modules/data_access/memberships/memberships.py
+++ b/infra/terraform/modules/data_access/memberships/memberships.py
@@ -63,23 +63,23 @@ def attach_bucket_policy(event, context):
     retry(fn)
 
 
-def retry(fn, attempts=5, delay=0.150):
-    tries = attempts
-    while tries > 0:
+def retry(fn, max_attempts=5, delay=0.150):
+    attempts = 0
+    while True:
         try:
             fn()
             break
         except botocore.exceptions.ClientError as error:
-            if error.response["Error"]["Code"] != "NoSuchEntity":
-                # Only retry on boto's ClientError/NoSuchEntity error
-                raise
-
-            tries -= 1
-            LOG.warning(
-                "error while attaching role to policy: {}.".format(error))
-            if tries > 0:
-                LOG.warning("Retry in {}s...".format(delay))
-                time.sleep(delay)
+            # Only retry on boto's ClientError/NoSuchEntity error
+            if error.response["Error"]["Code"] == "NoSuchEntity":
+                LOG.warning(
+                    "error while attaching role to policy: {}.".format(error))
+                attempts += 1
+                if attempts < max_attempts:
+                    LOG.warning("Retry in {}s...".format(delay))
+                    time.sleep(delay)
+                else:
+                    raise
             else:
                 raise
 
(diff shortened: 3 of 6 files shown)


infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -52,10 +53,30 @@ def attach_bucket_policy(event, context):
...
+                raise
+
+            tries -= 1
+            LOG.warning(
+                "error while attaching role to policy: {}.".format(error))
+            if tries > 0:
+                LOG.warning("Retry in {}s...".format(RETRY_DELAY))
+                time.sleep(RETRY_DELAY)
+            else:
+                raise
```
andyhd: This function is getting a bit long, and the code is operating at different levels of abstraction. This should be refactored to something like:
```python
retry(attach_role_policy(policy_arn, role), attempts=2)
```

andyhd: Also, there are a number of Python libraries which provide retry functionality like [retry](https://pypi.python.org/pypi/retry) and [retry-decorator](https://pypi.python.org/pypi/retry_decorator/1.0.0)

infra/terraform/modules/data_access/lambda_memberships.tf
```diff
@@ -5,8 +5,7 @@ resource "null_resource" "memberships_install_deps" {
     }
 
     triggers {
-        build_sh_sha = "${sha256(file("${path.module}/memberships/build.sh"))}"
-        requirements_sha = "${sha256(file("${path.module}/memberships/requirements.txt"))}"
+        force_rebuild = "${timestamp()}"
```
andyhd: Is this change connected to the retry functionality?

infra/terraform/modules/data_access/memberships/.gitignore
```diff
@@ -0,0 +1,4 @@
+contextlib2*
+naming*
+raven*
+sentry*
```
andyhd: Is this change connected to the retry functionality?

infra/terraform/modules/data_access/lambda_memberships.tf
xoen: This is a fix to something I noticed while working on it. Basically the building of the lambda function is not triggered on other devs machines because these `triggers`' attributes are kept in the TF remote store.

With hindsight I should have create a separate PR but I expected this PR to be merged sooner and that's why I included this small/kind-of-related fix here.

infra/terraform/modules/data_access/memberships/memberships.py
xoen: I moved the retry logic in a function. I kept it simple and avoided weird things like annotations, etc...`retry` is just a function which takes the function to retry.

xoen:
@andyhd I [extracted the retry logic](https://github.com/ministryofjustice/analytics-platform-ops/pull/89/commits/8ac4cafcfefc1bd656de2d49760628c7c5a3b13c). I tried it few times in the dev environment and it works. Interestingly to see how sometimes the policy is created very quickly requiring no retry, and how sometimes it takes a bit more time than the usual. Retry with these params (`attempts=5`, `delay=0.150`) should be a good starting point and reduce the Sentry false positives.

Let me know if can be merged :)

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -52,10 +53,35 @@ def attach_bucket_policy(event, context):
...
+            if error.response["Error"]["Code"] != "NoSuchEntity":
+                # Only retry on boto's ClientError/NoSuchEntity error
+                raise
+
+            tries -= 1
+            LOG.warning(
+                "error while attaching role to policy: {}.".format(error))
+            if tries > 0:
+                LOG.warning("Retry in {}s...".format(delay))
+                time.sleep(delay)
```
andyhd: If you `continue` here, you can remove the `break` from inside the `try` clause and, combined with inverting the logic of the error code check, you can remove one of the `raise` statements by just falling through both `if` statements, eg:
```python
while tries > 0:
    try:
        fn()
    except ... as error:
        if error... == 'NoSuchEntity':
            tries -= 1
            if tries > 0:
                time.sleep(delay)
                continue
        raise
```

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -52,10 +53,35 @@ def attach_bucket_policy(event, context):
...
+            RoleName=naming.role_name(username),
+            PolicyArn=policy_arn(team_slug, policy_type),
+        )
+
+    retry(fn)
+
+
+def retry(fn, attempts=5, delay=0.150):
+    tries = attempts
+    while tries > 0:
```
andyhd: This test is applied twice. It doesn't need to be here. This can be `while True` (unless you want to call `retry` with `attempts=0`, which makes no sense).

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -52,10 +53,35 @@ def attach_bucket_policy(event, context):
...
+        try:
+            fn()
+            break
+        except botocore.exceptions.ClientError as error:
+            if error.response["Error"]["Code"] != "NoSuchEntity":
+                # Only retry on boto's ClientError/NoSuchEntity error
+                raise
+
+            tries -= 1
+            LOG.warning(
```
xoen: 👍

infra/terraform/modules/data_access/memberships/memberships.py
```diff
@@ -52,10 +53,35 @@ def attach_bucket_policy(event, context):
...
+                "error while attaching role to policy: {}.".format(error))
+            if tries > 0:
```
xoen: Yeah, I passed through my mind as well. I like it more as well 👍

infra/terraform/modules/data_access/memberships/memberships.py
xoen: 👍

infra/terraform/modules/data_access/memberships/memberships.py
xoen: I moved to a `while True` loop so now I need the `break`, I'll push it soon so you can have another look 👍

andyhd: I checked, you don't need `break`:
```python
>>> x = 0
>>> while True:
...     try:
...         raise Exception('foo')
...     except Exception as e:
...         x += 1
...         if x < 5:
...             print('retry', x)
...             continue
...         raise
...
retry 1
retry 2
retry 3
retry 4
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
Exception: foo
```

infra/terraform/modules/data_access/memberships/memberships.py
xoen: Second attempt @andyhd (no pun intended 😹 ): https://github.com/ministryofjustice/analytics-platform-ops/pull/89/commits/c6e4b71da11bea7a4761cf836a28b6e5dda93d71

infra/terraform/modules/data_access/memberships/memberships.py
xoen: But what if never raise an exception? That's why there is a break after the fn(), for the happy path.
Have another look, I may miss something here.

xoen:
Thanks for your review @andyhd 👍

infra/terraform/modules/data_access/memberships/memberships.py
andyhd: Good point, I missed that 👍

PR #90, opened 2017-07-14
Add aws_account_logging module, S3 logs to elasticsearch lambda function, and consolidate S3 logs into a single bucket

kerin:
## What

This PR adds:

- a new Terraform module for AWS account-level logging (S3 access logs and in the future CloudTrail logs)
- a Lambda function to parse S3 logs on creation and insert them into Elasticsearch
- Elasticsearch document template and geoip ingest filter for S3 logs

It also replaces per-environment S3 logs buckets with a single global logs bucket for all environments.

## How to review

1. `terraform plan && terraform apply` in dev env
2. Access data in an S3 bucket configured for logging
3. Check Kibana for access log entry


charts/fluentd/templates/daemonset.yaml
```diff
@@ -31,6 +31,26 @@ spec:
               secretKeyRef:
                 name: {{ template "fullname" . }}
                 key: elasticsearch_port
+          - name:  FLUENT_ELASTICSEARCH_SCHEME
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "fullname" . }}
+                key: elasticsearch_scheme
+          {{ if .Values.elasticsearch.username }}
```
xoen: Stupid question: Why do we need to check if these values are present? I didn't check them in the `webapp` helm chart for example.

infra/terraform/modules/aws_account_logging/lambda.tf
```diff
@@ -0,0 +1,62 @@
+# Install dependencies
+resource "null_resource" "s3logs_install_deps" {
+    provisioner "local-exec" {
+        command = "${path.module}/s3logs/build.sh"
+    }
+
+    triggers {
```
xoen: As discussed earlier, `triggers` don't seem to work as we thought so let's just replace force the rebuild all the times until we find a better solution: https://github.com/ministryofjustice/analytics-platform-ops/pull/89/commits/19046f0c10da4cde551bdc6f0847e332b0d92a7d

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
```diff
@@ -0,0 +1,126 @@
...
+        'domain': 'localhost',
+        'port': '9200',
+        'index_prefix': 's3logs',
+        'doctype': 's3-access-log',
+        'params': 'pipeline=s3logs-geoip'
+    }
+
+    for key, value in env.items():
+        if key.startswith('ES_'):
+            values[key[3:].lower()] = value
```
xoen: I would personally be more explicit and just get the ElasticSearch params directly without checking/manipulating the environment variables names. This is a bit "meta" but it works.

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
```diff
@@ -0,0 +1,126 @@
...
+
+    return '{scheme}://{domain}:{port}/{index}/{doctype}?{params}'.format(
+        index=elasticsearch_url_index(values['index_prefix']),
+        **values)
```
xoen: Again, I would personally be more explicit in this case, you could probably replace the whole method (including defaults and for) with the following statement:

```python
return '{scheme}://{domain}:{port}/{index}/{doctype}?{params}'.format(
    scheme=os.environ.get('ES_SCHEME', 'http'),
    domain=os.environ.get('ES_DOMAIN', 'localhost'),
    port=os.environ.get('ES_PORT', '9200'),
    index=elasticsearch_url_index(values['index_prefix'],
    doctype='s3-access-log',
    params: 'pipeline=s3logs-geoip',
)
```

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
```diff
@@ -0,0 +1,126 @@
...
+
+    log_data = log_file['Body']._raw_stream
+    reader = codecs.getreader('utf-8')(log_data)
+
+    for line in reader:
+
+        line = line.strip()
+
+        if line:
+            yield line
```
xoen: nice one.

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
```diff
@@ -0,0 +1,126 @@
...
+
+    LOG.debug('Parsing log entry: {}'.format(entry))
+
+    log = s3_parser(entry)
+
+    log['time_received'] = log['time_received_utc_isoformat']
+
+    for key, value in dict(log).items():
+
+        if key.startswith('time_received_'):
```
xoen: Do we expect many keys with this prefix or it's just to remove the old key `time_received_utc_isoformat`? It we only want to delete this key it would be better to do so before the `for` loop.

charts/fluentd/templates/daemonset.yaml
kerin: This is to ensure that in cases where a username/password isn't required it doesn't result in an empty string being used, rather than username/password not being sent at all. This might not actually be necessary though - I'll check

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
kerin: This is @andyhd's doing 😬

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
kerin: This is also @andyhd's doing 😬

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
kerin: Yep, there's actually 6 datetime fields in the parsed object, so this strips all the others out.

xoen: 👍

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
xoen: It's just that if you jump in the code and see that `values[key[3:].lower()] = ...` is a bit opaque.

But as I said, it works :)

xoen:
As far as I can tell looks good 👍

charts/fluentd/templates/daemonset.yaml
kerin: Having had a look at this it seems like there's no easy way to handle optional values in Secrets - because values must be b64 encoded, null values cause syntax errors, so it seems like wrapping them in `{{ if }}` blocks is the simplest option.

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
andyhd: I replaced this code to make the default arguments more obvious and explicit

andyhd: Using `get` to retrieve values from the `os.environ` dict in this way is essentially performing multiple operations in a single long line of code.

infra/terraform/modules/aws_account_logging/s3logs/s3_to_es.py
andyhd: I agree, processing the `ES_` prefixed dict keys should be factored out

PR #98, opened 2017-08-09
Fix cloudtrail logs geoip pipeline and errors with digest files

kerin:
- Cloudtrail logs `sourceIPAddress` sometimes contains a hostname rather than an IP address, which causes errors with Elasticsearch, so don't use the pipeline in those cases
  - Exclude Cloudtrail digest files from lambda function trigger

### DIFF
--- a/infra/terraform/modules/aws_account_logging/cloudtrail/cloudtrail_to_es.py
+++ b/infra/terraform/modules/aws_account_logging/cloudtrail/cloudtrail_to_es.py
@@ -8,8 +8,8 @@ import ipaddress
 import boto3
 from botocore.client import Config
 import certifi
-import elasticsearch
 from elasticsearch import Elasticsearch
+import elasticsearch.helpers
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


infra/terraform/modules/aws_account_logging/cloudtrail/cloudtrail_to_es.py
```diff
@@ -47,28 +49,41 @@ def post_to_elasticsearch(doc):
...
 
+def get_pipeline(doc):
+    # sourceIPAddress may not be present, or may contain a hostname,
+    # which causes an ElasticSearch error - so only specify the geoip
+    # pipeline if an actual IP address is present
+    try:
+        ipaddress.ip_address(doc['sourceIPAddress'])
+        return 'cloudtrail-geoip'
+    except (KeyError, ValueError):
+        return None
```
andyhd: It's better to minimize the code inside a `try` block. You can also avoid the later check by returning an empty string instead of `None` if there is an exception. And if you use `doc.get()` you only need to catch `ValueError`:
```python
try:
    ipaddress.ip_address(doc.get('sourceIPAddress'))

except ValueError:
    return 'pipeline=cloudtrail-geoip'

return ''
```

andyhd: I'd be tempted to separate the test for a valid IP address from setting the URL parameter, for a more self documenting approach:
```python
def is_ip_address(source_ip_address):
    try:
        ipaddress.ip_address(source_ip_address)

    except ValueError:
        return False

    return True
```
and then
```python
params = {}

if is_ip_address(doc.get('sourceIPAddress')):
    params['pipeline'] = 'cloudtrail-geoip'

requests.post(elasticsearch_url(os.environ), params=params, ...)
```

PR #99, opened 2017-08-10
Remove jenkinsfiles used by control panel

andyhd:
* Deleted jenkinsfiles and supporting scripts - these are now in
https://github.com/ministryofjustice/analytics-platform-control-panel


xoen:
Nice 👍

PR #102, opened 2017-08-10
Remove jupyter notebook kubernetes templates

andyhd:
These should be in ministryofjustice/analytics-platform-helm-charts.

This old Jupyter setup is not used and when we come to implementing it, it should not be in this repository.


xoen:
Looks good to me 👍

PR #103, opened 2017-08-10
Remove documentation of removed jenkinsfiles, charts and config

andyhd:


xoen:
LGTM 👍

xoen:
I think @kerin 's idea is good 👍

andyhd:
Is that ok?

PR #109, opened 2017-09-27
Fix control_panel_api IAM role

xoen:
### What
The `control_panel_api` IAM role wasn't assumed by the control panel API
instances. The problem was a missing statement with the k8s workers role
ARN as principal.

This is necessary in order for kube2iam to work as expected (see below)

### `sts:AssumeRole` required by kube2iam

See kube2iam documentation on assume role/trust relationship the IAM role needs:
https://github.com/jtblin/kube2iam#iam-roles

### Ticket:

https://trello.com/c/hSOlOUmE/391-cp-api-bug-unable-to-locate-aws-credentials


xoen:
I'll take you're happy with this @kerin right? Not urgent, but let me know so I can close the case :)

PR #110, opened 2017-09-28
Alpha: Uncomment 'control_panel_api' module

xoen:
### What

This will create the CP-API DB and its IAM role.

I also added the DB host and the IAM role to the outputs.

### Ticket

Part of: https://trello.com/c/8Y086TqG/407-cp-api-deploy-to-alpha


xoen:
Cheers!!

PR #111, opened 2017-09-28
Updated cpanel alpha credentials

xoen:
Made alpha DB password comply with password rules.


xoen:
FYI: @kerin this is the password thing we already discussed offline. I'm merging it.

PR #113, opened 2017-12-06
Lambda functions zips depends on script which install dependences

xoen:
### What

Two of the lambda functions' zips which must include some dependences
were not depending on the `null_resource` which install them.

### Why
This can *potentially* cause the installation to happen *after* the zip
is created, which therefore will miss these dependences.

### Ticket
Possible solution to https://trello.com/c/MpBnZkQx/548-lambda-function-i-o-error-5


xoen:
FYI: I'm not applying this yet as `terraform plan` list other non-related changes (in `dev` at least) so I may want to apply this with someone else checking - just to be on the safe side.

PR #123, opened 2018-01-25
Control Panel IAM role can read/write roles' inline policies

xoen:
This is necessary for IAM roles revamp to work:

https://github.com/ministryofjustice/analytics-platform-control-panel/pull/88

After this changes CP-API will need to read and write IAM roles' inline policies which requires `iam:GetRolePolicy` and `iam:PutRolePolicy`.

### Ticket
Part of https://trello.com/c/2nuxrwiq/586-5-iam-policies-change-from-1-per-bucket-to-1-per-user


xoen:
FYI: Terraform-Applied in `dev` and tested from a CP-API pod and it works as expected.

xoen:
FYI: Applied in `alpha`. Merging.

PR #124, opened 2018-01-29
(TEMPORARY) CP-API role permission to read/update users roles trust policies

xoen:
### What
The [migration which adds the OIDC statement to the users' roles trust policies](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/90) will need the `iam:GetRole` and `iam:UpdateAssumeRolePolicy` permissions to read/update them.

**NOTE**: This will be needed *only* before running the migration.

After running the migration we can remove them.
In fact, there is no point in merging this, better just apply off this branch.

### Part of ticket
https://trello.com/c/Vy7ImBcn/602-4-sso-aws-federated-login-user-roles-migration


xoen:
FYI: Applied in `dev` but **NOT** in `alpha`.

xoen:
[(4) SSO/AWS federated login user roles migration](https://trello.com/c/Vy7ImBcn/602-4-sso-aws-federated-login-user-roles-migration)

PR #125, opened 2018-02-01
Add s3:PutBucketTagging

jmoz:
## What

This enables the new call to put bucket tagging on a 'data warehouse' bucket.

Related to PR on api https://github.com/ministryofjustice/analytics-platform-control-panel/pull/93


jmoz:
Tested with terraform apply

PR #128, opened 2018-03-14
Init-User Chart Redeploy script

djsd123:
## What    

Trello: https://trello.com/c/1h5t0me5

Simple script to redeploy the init-user helm chart.

Will delete the rstudio and jupyter releases belonging to the user as the resources for those releases will be removed by deleting the init-user release.

## How to review

1. `kubectl config use-context < Your Dev cluster context >
2.  Ensure the directory you're in has both [Helm Charts](https://github.com/ministryofjustice/analytics-platform-helm-charts) and [Chart Config](https://github.com/ministryofjustice/analytics-platform-config) repos cloned to it.
3. Run the script `./helm_redeploy_init-user.sh`

### DIFF
--- a/scripts/helm_redeploy_init-user.sh
+++ b/scripts/helm_redeploy_init-user.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env sh
+
+ENVIRONMENT=""
+
+while true; do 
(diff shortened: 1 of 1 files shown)


djsd123:
@andyhd because `sh` arrays if you can call them that do not hold state.  So after the `del --purge`, there's no data to re-deploy the releases.

PR #130, opened 2018-03-26
Add permission to list IAM policies to control_panel IAM role

xoen:
### What
Added permission to `iam:ListPolicies` to `control_panel_api` IAM role

### Why
This is needed by the [migration](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/files#diff-7c3ac2e9db789dd8ece232021649f51aR38) which reads IAM policies and
creates the corresponding S3 buckets in the Control Panel API DB.

### See
- [migration PR](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109)
- [IAM documentation at iam.cloudonaut.io](https://iam.cloudonaut.io/reference/iam.html)

### TODO
- [X] Apply in `dev`
- [x] Test in `dev` to see if migration can read IAM policies
- [ ] Apply in `alpha`

### Part of ticket
https://trello.com/c/Me2RD6ek/696-5-script-that-copies-data-access-permissions-from-existing-lambda-created-policies-to-to-s3-access-inline-policy-and-adds-these

### DIFF
--- a/infra/terraform/modules/control_panel_api/role.tf
+++ b/infra/terraform/modules/control_panel_api/role.tf
@@ -135,6 +135,16 @@ resource "aws_iam_policy" "control_panel_api" {
       ]
     },
     {
+      "Sid": "CanListPoliciesForLambdaMigration",
+      "Effect": "Allow",
+      "Action": [
+        "iam:ListPolicies"
+      ],
+      "Resource": [
+        "arn:aws:iam::${var.account_id}:*"
+      ]
+    },
+    {
       "Sid": "TemporaryForOIDCMigration",
       "Effect": "Allow",
       "Action": [


infra/terraform/modules/control_panel_api/role.tf
```diff
@@ -134,6 +134,16 @@ resource "aws_iam_policy" "control_panel_api" {
...
       ]
     },
+    {
+      "Sid": "CanListPoliciesForLambdaMigration",
+      "Effect": "Allow",
+      "Action": [
+        "iam:ListPolicies"
+      ],
+      "Resource": [
+        "arn:aws:iam::${var.account_id}:*"
```
djsd123: Allowed to list policies on any resource?  Is that what you meant to do?

xoen: As far as I know this is to list IAM policies in your account. The action to list policies attached to another IAM resource is different, something like `iam:ListAttachedRolePolicies`.

But in this case I'm interested in listing all the IAM policies in our account, if you [see the migration](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/109/files#diff-7c3ac2e9db789dd8ece232021649f51aR36) it's starting by reading all the IAM policies, etc...

xoen:
FYI: I ran the migrations in `dev` and I did't get any permissions problems so merging this and applying in `alpha` soon (today or tomorrow morning)

PR #131, opened 2018-04-04
Removed modules related to data-access lambdas

xoen:
### What
The following modules will not be required anymore once users will
start to use the Control Panel to grant/revoke access to data:
- webhooks_api
- notifications
- data_access

### Ticket
Ticket: https://trello.com/c/Ex7mKhYS/712-turn-off-lambda-functions

### NOTE
**Don't apply/merge this** until we didn't migrate the buckets/memberships from AWS to CP-API and we're happy for users to use UI.


xoen:
@kerin This will have to wait to be applied/merged but have a look anyway to see if it makes sense to you.

PR #133, opened 2018-04-20
Adds NOMIS/OASYS upload user and S3 uploads bucket

kerin:
## Data upload user Terraform module

Adds data_upload_user Terraform module

This module creates required AWS resources for S3 upload system users. Each invocation of this module will create:

- An IAM user
- An AWS access key for that user
- An IAM policy allowing write-only access to a specific org and system path in the specified S3 bucket

The intended use is for this module to be invoked once per system user.

IAM users are created under a `/uploaders/$org_name` path (e.g. `/uploaders/hmpps`), and IAM policies are created under a `/uploaders/$org_name/$system_name` path (e.g. `/uploaders/hmpps/nomis_oasys`

S3 access is granted to specific paths only, with a similar structure as above - `BUCKET/uploaders/$org_name/$system_name`

## Data upload S3 bucket

Adds S3 bucket for system uploads

Single bucket for system data uploads. Uses default object encryption with customer-owned KMS keys.

The use of a default bucket encryption policy requires a recent version of the Terraform AWS provider, so a version requirement has been added.


infra/terraform/modules/data_upload_user/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "system_user_access_key_id" {
```
xoen: Minor point here. I'd keep these output names shorter and just rename them:
- `system_user_access_key_id` => `access_key_id`
- `system_user_access_key_secret` => `access_key_secret`

The fact these are for a user is kind of implied by the fact this is in the 'data_upload_**user**' module.

infra/terraform/global/s3.tf
```diff
@@ -15,3 +15,27 @@ resource "aws_s3_bucket" "terraform_state" {
         enabled = true
     }
 }
+
+resource "aws_kms_key" "s3_uploads" {
+    description = "Key for upload object encryption"
+    deletion_window_in_days = 10
```
xoen: I read the terraform documentation on this `deletion_window_in_days` argument.

> Duration in days after which the key is deleted after destruction of the resource, must be between 7 and 30 days. [...]

So we may delete the terraform resource but the KMS key sticks around for another 10 days in AWS? Any reasons why it's not deleted straight away? (maybe because if you accidentally delete it you will "never" be able to decrypt your data anymore? so they keep it around just-in-case?)

infra/terraform/global/s3.tf
```diff
@@ -15,3 +15,27 @@ resource "aws_s3_bucket" "terraform_state" {
...
+    acl = "private"
+
+    versioning {
+        enabled = true
+    }
+
+    server_side_encryption_configuration {
+        rule {
+            apply_server_side_encryption_by_default {
+                kms_master_key_id = "${aws_kms_key.s3_uploads.arn}"
```
xoen: Another KMS stupid question: How does this work? If S3 is going to use this KMS key transparently on the server, this means S3 can access this key, do we need to explicitly give S3 permission to this key or something (via IAM)?

infra/terraform/global/variables.tf
```diff
@@ -11,3 +11,5 @@ variable "es_password" {}
 variable "es_scheme" {
   default = "https"
 }
+
+variable "uploads_bucket_name" {}
```
xoen: We may have a default value here. Unless we think the bucket name itself is somehow sensitive (don't think so).

infra/terraform/modules/data_upload_user/variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "upload_bucket_arn" {}
+variable "org_name" {}
+variable "system_name" {}
```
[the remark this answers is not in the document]
xoen: Right. I see why you called this `system_name`, because this is the username for a script uploading the files 👍

kerin:
@RobinL no that's a fair point, and @xoen said the same. The rationale was that the existing process uses a single script and set of credentials for both systems, but on reflection I doubt there's any issue with us separating them out, so I've changed this to create a separate set of credentials for each system.

infra/terraform/modules/data_upload_user/iam.tf
```diff
@@ -0,0 +1,26 @@
+resource "aws_iam_user" "system" {
+  name = "${var.system_name}"
+  path = "/uploaders/${var.org_name}/"
```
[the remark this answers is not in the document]
kerin: That's because the name attribute of the IAM user is the system name, so the final ARN will end with `:role/uploaders/hmpps/nomis`, whereas the policy's name reflects the permissions granted, so the ARN there will end with `:policy/uploaders/hmpps/nomis/s3_upload_writeonly`

infra/terraform/global/s3.tf
kerin: Yep, exactly - you can't delete a KMS key immediately, and 7 days in the minimum.

infra/terraform/modules/data_upload_user/outputs.tf
kerin: good point - changed

infra/terraform/modules/data_upload_user/iam.tf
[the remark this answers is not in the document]
xoen: Oh, right, sorry that makes perfectly sense 👍

infra/terraform/global/s3.tf
kerin: @xoen @RobinL FYI, I tried using customer-managed KMS encryption keys on this bucket as per AWS's best practice advice, but this resulted in permission errors. Documentation on required permissions for this are unclear or hard to find, so I've changed back to S3-managed keys for now.

PR #135, opened 2018-04-30
Set Image Tag

djsd123:
## What

Script to modify image tags.  Useful for rolling out
versions or just serving as an example. As is will perform the migration on
rstudio.

Gated with `dry-run` flag for no-op runs to what will be changed if
run.

Title the PR to complete the sentence: "Merging this PR will ..."

## How to review

Execute the script and answer `y` to invoke a dry run. You will see output for deployments that will be affected. i.e. __Does not match __ the image defined in the script.

In dev, that will be my deployment

### DIFF
--- a/scripts/set_deployment_image_tag.sh
+++ b/scripts/set_deployment_image_tag.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Generic script that sets the image+tag of a container in a deployment's pod spec in user namespaces
+# As is.  Will set deployment's "r-studio-server" container's image tag to "v1.3.2" in all user namespaces
+
+set -e
+
+DRY_RUN=true
+
+while true; do
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
+done
+
+APP=rstudio
+TAG=v1.3.2
+POD_CONTAINER=r-studio-server
+NAMESPACES=$(kubectl get ns | grep user- | grep -v user-init-platform | cut -f1 -d' ')
+
+
+
+for ns in $NAMESPACES; do
+
+	kubectl set image deployments -l app=$APP $POD_CONTAINER=quay.io/mojanalytics/$APP:$TAG \
+	--namespace $ns \
+	--dry-run=$DRY_RUN
+
+done


scripts/set_deployment_image_tag.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Generic script that sets the image+tag of a container in a deployment's pod spec in user namespaces
+# As is.  Will set deployment's "r-studio-server" container's image tag to "v1.3.2" in all user namespaces
+
+set -e
+
+DRY_RUN=true
+
+while true; do
```
xoen: Fancy dialog, uh :P?

scripts/set_deployment_image_tag.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
xoen: I wonder if this should be [with the helm chart](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/master/charts/rstudio/migrations/20180411-1155_increase_cookie_maxage.sh) or not.

scripts/set_deployment_image_tag.sh
```diff
@@ -0,0 +1,32 @@
...
+while true; do
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
+done
+
+APP=rstudio
```
xoen: If this is not a migration but a script will run often it may be worth having these 3 constants read from environment variables with a good default.

djsd123:
I agree.  This one should be more generic.  I will make the suggested changes.

scripts/set_deployment_image_tag.sh
```diff
@@ -0,0 +1,32 @@
...
+TAG=v1.3.2
```
xoen: I saw this in the helm-charts repo as well, I think you were right about having this in the `*-ops` repo until we find a better place. 

I would also add a param for the `$IMAGE`, as I've [done here](https://github.com/ministryofjustice/analytics-platform-helm-charts/commit/596ba662098914b55a0439dae91f2b7553f14827#diff-fbf8fa4423f8d08f8e559443cea7f5df), so we can potentially update the other container image (e.g. the proxy) and probably delete the script in the helm-charts repo as it's a duplication now :)

scripts/set_deployment_image_tag.sh
r4vi: not saying change it but have you seen the `dialog` command? 
`dialog --yesno "Is this a dry run?" 6 25; echo $?`

xoen: ✨  - didn't know it. curses galore :D 

I wonder how that would work in a non-interactive run, I played with [`--defaultno --sleep 5`](https://linux.die.net/man/1/dialog) but it doesn't seem to do what I would expect (default to "No" after 5 seconds unless I select something else before) 🤷‍♂️  - nice trick tho.

r4vi: yeah it's a bit too fancy, I actually prefer this `while true` but I guess you can do `--timeout 5` instead of sleep, but timeout always exits with code 255

PR #136, opened 2018-05-14
Update grafana config for platform users

axemonkey:
I'm assuming it's normal for an update to this to change pretty much every line...? Maybe just because it hasn't been updated in 7 months...?

This is just pasted out of the Grafana UI as it stands right now.

### DIFF
--- a/config/grafana/dashboards/platform-users.json
+++ b/config/grafana/dashboards/platform-users.json
@@ -1,500 +1,499 @@
 {
   "annotations": {
-    "list": []
+    "list": [
+      {
+        "$$hashKey": "object:79",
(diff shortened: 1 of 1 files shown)


djsd123:
@axemonkey, Yeah I see what you mean.  Hiding the whitespace in the diff makes it easier to see the changes.  I've got no beef with the changes. @kerin?

kerin:
Yep looks fine - bear in mind that this stuff was added to the repo just so we had it stored _somewhere_, and was always just meant to be a copy of what's going on in the Grafana instance, so just copying and pasting it into the repo make sense 👍

PR #139, opened 2018-05-16
Lambda Module

djsd123:
## What

Done as part of [Trello](https://trello.com/c/DAeTcJoT)

Module for provisioning Lambda functions

Default runtime `Golang`

IAM policies to be passed in at module invocation

## How to review

I've already deployed this but to kind of test do:

1. `git checkout ebs-backups`
2. `terraform workspace new test-lambda`
3. `terraform init`
4. `terraform plan -target=module.kubernetes_etcd_ebs_snapshot`
5. See: [README](https://github.com/ministryofjustice/analytics-platform-ops/pull/139/files#diff-4a0623065a272190f6672c2feb083203) for an example of how to implement

### DIFF
--- a/infra/terraform/modules/lambda_mgmt/CHANGELOG.md
+++ b/infra/terraform/modules/lambda_mgmt/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [0.1.0] - 2018-05-16
+### Lambda Management Module
+- Initial Commit
--- a/infra/terraform/modules/lambda_mgmt/README.md
+++ b/infra/terraform/modules/lambda_mgmt/README.md
@@ -1,0 +1,62 @@
+## Lambda Management 
+====================
+
+Used to provision lambda functions
+
+Module Input Variables
+----------------------
+
+- `lambda_function_name` - Name for Lambda function
+- `lambda_runtime` - A [valid](http://docs.aws.amazon.com/cli/latest/reference/lambda/create-function.html#options) Lambda runtime environment. Defaults to `go1.x`
+- `zipfile` - Path to zip archive containing Lambda function
+- `handler` - The entrypoint into your Lambda function, in the form of `filename.function_name`. For `Golang` this must match `lambda_function_name`
+- `schedule_expression` - A [valid rate or cron expression](http://docs.aws.amazon.com/lambda/latest/dg/tutorial-scheduled-events-schedule-expressions.html)
+- `source_code_hash` - The base64 encoded sha256 hash of the archive file - see TF [archive file provider](https://www.terraform.io/docs/providers/archive/d/archive_file.html)
+- `timeout` - (optional) The amount of time your Lambda Function has to run in seconds. Defaults to 3. See [Limits](https://docs.aws.amazon.com/lambda/latest/dg/limits.html)
+- `enabled` - (optional) Boolean expression. If false, the lambda function and the cloudwatch schedule are not set. Defaults to `true`.
+- `env_key_*` - (optional) The key of an environment variable to set for your lambda function
+- `env_value_*` - (optional) The value of an environment variable to set for your lambda function
+- `lambda_policy` - The IAM policy document.  Usually JSON 
+
+Usage 
+-----
+
+The example below provisions a lambda function that manages ebs snapshots
+
+```
+data "template_file" "lambda_create_snapshot_policy" {
+  template = "${file("assets/create_etcd_ebs_snapshot/lambda_create_snapshot_policy.json")}"
+}
+
+data "archive_file" "kubernetes_etcd_ebs_snapshot_code" {
+  source_file = "assets/create_etcd_ebs_snapshot/create_etcd_ebs_snapshot"
+  output_path = "assets/create_etcd_ebs_snapshot/create_etcd_ebs_snapshot.zip"
+  type        = "zip"
+}
+
+module "kubernetes_etcd_ebs_snapshot" {
+  source               = "../modules/lambda_mgmt"
+  lambda_function_name = "create_etcd_ebs_snapshot"
+  zipfile              = "assets/create_etcd_ebs_snapshot/create_etcd_ebs_snapshot.zip"
+  handler              = "create_etcd_ebs_snapshot"
+  source_code_hash     = "${data.archive_file.kubernetes_etcd_ebs_snapshot_code.output_base64sha256}"
+  env_key_1            = "INSTANCE_TAG_KEY"
+  env_value_1          = "k8s.io/role/master"
+  env_key_2            = "INSTANCE_TAG_VALUE"
+  env_value_2          = "1"
+  lamda_policy         = "${data.template_file.lambda_create_snapshot_policy.rendered}"
+}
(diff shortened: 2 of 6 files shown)


infra/terraform/modules/lambda_mgmt/README.md
```diff
@@ -0,0 +1,60 @@
...
+}
+
+data "archive_file" "kubernetes_etcd_ebs_snapshot_code" {
+  source_file = "assets/create_etcd_ebs_snapshot/create_etcd_ebs_snapshot"
+  output_path = "assets/create_etcd_ebs_snapshot/create_etcd_ebs_snapshot.zip"
+  type        = "zip"
+}
+
+module "kubernetes_etcd_ebs_snapshot" {
+  source               = "github.com/ministryofjustice/analytics-platform-ops/infra/terraform/modules/lambda_mgmt"
```
xoen: Interesting. So as we pass the absolute path, this is effectively an "external" terraform module?

infra/terraform/modules/lambda_mgmt/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+- `lambda_function_name` - Name for Lambda function
+- `lambda_runtime` - A [valid](http://docs.aws.amazon.com/cli/latest/reference/lambda/create-function.html#options) Lambda runtime environment. Defaults to `go1.x`
+- `zipfile` - Path to zip archive containing Lambda function
+- `handler` - The entrypoint into your Lambda function, in the form of `filename.function_name`. For `Golang` this must match `lambda_function_name`
+- `schedule_expression` - A [valid rate or cron expression](http://docs.aws.amazon.com/lambda/latest/dg/tutorial-scheduled-events-schedule-expressions.html)
+- `source_code_hash` - The base64 encoded sha256 hash of the archive file - see TF [archive file provider](https://www.terraform.io/docs/providers/archive/d/archive_file.html)
+- `timeout` - (optional) The amount of time your Lambda Function has to run in seconds. Defaults to 3. See [Limits](https://docs.aws.amazon.com/lambda/latest/dg/limits.html)
+- `enabled` - (optional) Boolean expression. If false, the lambda function and the cloudwatch schedule are not set. Defaults to `true`.
+- `tag_key` - (optional) The key of the tag to filter resources on
```
xoen: I'm not 100% sure about this module (just started to have a look). Is the idea to have this to "provision" different lambda functions? But then why do we have these input variables specific of a certain lambda function (the ELB snapshot one)?

infra/terraform/modules/lambda_mgmt/lambda.tf
```diff
@@ -0,0 +1,27 @@
...
+  filename         = "${var.zipfile}"
+  handler          = "${var.handler}"
+  role             = "${aws_iam_role.lambda_role.arn}"
+  runtime          = "${var.lambda_runtime}"
+  source_code_hash = "${var.source_code_hash}"
+  count            = "${var.enabled}"
+  timeout          = "${var.timeout}"
+  environment {
+    variables {
+      TAG_KEY   = "${var.tag_key}"
```
xoen: See comment above about generic vs. specific lambda function input/environment variables.

infra/terraform/modules/lambda_mgmt/variables.tf
```diff
@@ -0,0 +1,45 @@
...
+variable "enabled" {
+  default = true
+}
+
+variable "timeout" {
+  default = 3
+}
+
+variable "schedule_expression" {
+  default = "rate(1 day)"
```
xoen: I suppose we don't care too much when the snapshots/invocation happen.

infra/terraform/modules/lambda_mgmt/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
```
xoen: 🏆

xoen:
As far as I can tell it looks alright to me. I'm not 100% sure about the utility of the module considering the variables set in the lambda resource are specific of the ELB snapshot lambda.

Maybe there is a way in terraform to have nested variables, so the module would read variables under `variables` and set them in the lambda, e.g. 

```
module "kubernetes_etcd_ebs_snapshot" {
  source               = "github.com/ministryofjustice/analytics-platform-ops/infra/terraform/modules/lambda_mgmt"
  variables = {
    tag_key = // ...
    tag_value = // ...
  }
  # ...
```

Or something like that, and these values under `variables` would end-up in the in the `aws_lambda_function` under `environment.variables`...

...BUT, I'm not sure you can do something like that in terraform (e.g. [there is a issue open](https://github.com/hashicorp/terraform/issues/2114)).

infra/terraform/modules/lambda_mgmt/README.md
djsd123: Oh sorry, that was just an example.  Typically from my experience modules always lived in their own repos.  I'll change to filesystem path.

infra/terraform/modules/lambda_mgmt/README.md
djsd123: Agreed.  Could be more generic.

infra/terraform/modules/lambda_mgmt/variables.tf
djsd123: At first, I didn't but I supposed 10 or 11 PM could be sensible. I'll leave this as a default and change at module invocation

infra/terraform/modules/lambda_mgmt/CHANGELOG.md
djsd123: Yay 💯

djsd123:
@xoen, not to rush mate but how do I move this along?  I've changed the names of the env variable keys.  I think the names of the env keys might indicate that the module performs operations specifically on `ec2` resources but the implementation of the module itself, I feel is fairly generic..

If anything, it might need more optional env vars?  However, I'd cross that bridge....

djsd123:
@xoen, @kerin, could you have another look, please!

infra/terraform/modules/lambda_mgmt/lambda.tf
```diff
@@ -0,0 +1,28 @@
...
+  handler          = "${var.handler}"
+  role             = "${aws_iam_role.lambda_role.arn}"
+  runtime          = "${var.lambda_runtime}"
+  source_code_hash = "${var.source_code_hash}"
+  count            = "${var.enabled}"
+  timeout          = "${var.timeout}"
+
+  environment {
+    variables {
+      "${var.env_key_1}" = "${var.env_value_1}"
```
xoen: uh, meta ✨

xoen: Finally what you said makes more sense. The reason why I wasn't "accepting" it was because for some reason I assumed that you couldn't control what was on the left of the `=` sign 🤷‍♂️ 

But this makes sense now, I mean, it's obviously a bit hacky but if it works it wouldn't change anything inside the lambda function 👍

infra/terraform/modules/lambda_mgmt/variables.tf
```diff
@@ -0,0 +1,54 @@
...
+}
+
+variable "source_code_hash" {}
+
+variable "env_key_1" {
```
xoen: Now that I see what you're doing here, it may make sense to call these `env_1_name`/`env_1_value`, `env_2_name`/`env_2_value`. But sorry, I know it's a bit annoying, so you're call.

djsd123: Sure.  Ok with me.

PR #140, opened 2018-05-23
Linting and validation of Terraform resources

kerin:
## What

- add circleCI job to validate and lint terraform code
- pass all existing terraform code through `terraform validate && terraform fmt`


kerin:
@djsd123 yeah - I wonder if maybe it'd be a good idea to have `terraform fmt` set up as a pre-commit hook, which I think is a common usage of `gofmt`?

djsd123:
👍

xoen:
@kerin sorry for the delay, I'll skim through the changes now to see if there is anything wrong (but in theory `terraform fmt` should change the semantic of the code)

Regarding the idea of a pre-commit hook it feels a bit magical to me, so personally I'd prefer CircleCI to show you a big red cross and you know that you forgot to run `terraform fmt`. My 2 cents of course.

.circleci/config.yml
```diff
@@ -0,0 +1,31 @@
+# based on: https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/.circleci/config.yml
```
xoen: I see that the [original also uses `tflint`](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/.circleci/config.yml#L25) which has additional tests (e.g. invalid EC2 instance type).

I wonder if it may be useful for as well.

xoen:
I skimmed through and all seems OK to me (as expected!). Nice one ✨

PR #142, opened 2018-05-23
Replace per-env terraform config with workspaces

kerin:
Terraform’s new(ish) workspace feature allows us to define terraform resource configuration once, and apply it to multiple conceptual environments, each with a different remote state behind them.

This means that applying a change to two environments changes from:

```
cd infra/terraform/environments/dev
terraform apply
cd ../alpha
terraform apply
```

to:

```
cd infra/terraform/platform
terraform workspace select dev
terraform apply -var-file=dev.tfvars
terraform workspace select alpha
terraform apply -var-file=alpha.tfvars
```

Crucially, the use of workspaces means that the same collection of terraform resources is used for both environments, so we no longer need to copy and paste changes between the two environments.

**Important**: enabling workspaces changes the remote state store behind an environment, so simply running `terraform apply` on a workspace-enabled version will create duplicates of all resources, name clashes from the existing state, dogs and cats living together, and more. So as and when this PR is merged, we must manually move the existing state to the new workspace-enabled state store using `terraform state` commands.


xoen:
Just starting to have a look at this (just read the PR description) and I have two comments to make:
1) Any way to avoid passing the variables explicitly when running `terraform apply`? e.g. to tell terraform where to find the variables for a given workspace. 
2) So migrate to workspaces involves manually change the TF state 😱?

Reason why I'm asking 1) is that I imagine passing that explicitly/manually is error prone and the risk is that someone run `terraform apply` passing tfvars for the wrong environment by accident 💥

andyhd:
Yeah, but you could do this to avoid that:
```sh
cd infra/terraform/platform
for ws in "dev" "alpha"; do
  terraform workspace select $ws
  terraform apply -var-file=$ws.tfvars
```

infra/terraform/platform/main.tf
```diff
@@ -1,55 +1,44 @@
...
-data "terraform_remote_state" "base" {
-  backend = "s3"
-
-  config {
-    bucket = "${var.terraform_bucket_name}"
-    region = "${var.region}"
-    key    = "${var.terraform_base_state_file}"
-  }
+  region  = "${var.region}"
+  version = "~> 1.19"
```
djsd123: 👍 💯

kerin:
@xoen unfortunately there's no way to have terraform automatically pick up a `.tfvars` file for a given workspace, so you have to manually specify it - there's an open issue for it somewhere in the Terraform github repo though. Obviously that is error prone, but we're working towards having this in an automated pipeline though, so there'd be a step like `terraform workspace select $env && terraform apply -var-file=${env}.tfvars`, which would mitigate that.

And yes, introducing workspaces on an existing project does mean that state needs to be manually migrated, because [terraform interpolates the workspace name into the state path](https://www.terraform.io/docs/backends/types/s3.html#key), but that's unavoidable unfortunately. In theory it'd just be case of using `terraform state mv` on the root object for an environment, but I need to investigate that further to be sure

xoen:
I've just noticed that the README needs some love ❤️, it would be good to update it so that it reflects the new way to operate.

infra/terraform/platform/main.tf
```diff
@@ -1,55 +1,44 @@
 terraform {
   backend "s3" {
     bucket = "terraform.analytics.justice.gov.uk"
-    key    = "dev/terraform.tfstate"
+    key    = "platform.tfstate"
```
xoen: So, just to state the obvious/double check: The TF state would be in a sub "directory" in the S3 bucket called after the workspace name right?

Is that what you meant by saying "copying the state"? As in copying this `(dev|alpha)/terraform.tfstate` to `(dev/alpha)/platform.tfstate`? (Or something like that)

infra/terraform/platform/main.tf
```diff
@@ -60,46 +49,46 @@ module "user_nfs_softnas" {
...
 }
 
 module "container_registry" {
-  source = "../../modules/container_registry"
-  env    = "${var.env}"
+  source = "../modules/container_registry"
+  env    = "${terraform.workspace}"
 }
 
 module "federated_identity" {
```
xoen: I just noticed that we still have the `modules/federated_identity/saml` directory, it may be a good opportunity to delete it.

I mean, it's still committed and we can get it back if we need to, but I don't think it's used anywhere.

xoen:
I may miss something but the path to the var file should include `vars/`, e.g.

```bash
cd infra/terraform/platform
terraform workspace select dev
terraform apply -var-file=vars/dev.tfvars
```

(In case that goes in the README)

infra/terraform/platform/main.tf
kerin: (Yep)[https://www.terraform.io/docs/backends/types/s3.html#key] - according to that, the actual state object for `dev` will be at `/:env/dev/platform.tfstate`

"Copying the state" refers to copying the data in the existing state object (`dev/terraform.tfstate`) to the new one (`/:env/dev/platform.tfstate`) - although it turns out that `terraform state` has no `copy` command, only `mv`, so we'll need to either move the state data, or import existing resources into the new state object.

infra/terraform/platform/main.tf
```diff
@@ -1,55 +1,44 @@
...
+  env = "${terraform.workspace}"
 }
 
 module "user_nfs_softnas" {
-  source = "../../modules/user_nfs_softnas"
+  source = "../modules/user_nfs_softnas"
 
   num_instances             = 1
-  env                       = "${var.env}"
+  env                       = "${terraform.workspace}"
```
xoen: Do modules have a access to `terraform.workspace` or it's not visible from within the module? 

Don't know if it's considered a "bad practice" but if you can access `terraform.workspace` from inside a module maybe we can get rid of the input variable `env` we have in all these modules.

I don't know, maybe it's bad that you introduce this implicit dependency on something outside the module.

infra/terraform/platform/main.tf
```diff
@@ -1,55 +1,44 @@
...
 
-  env = "${var.env}"
+  env = "${terraform.workspace}"
 }
 
 module "user_nfs_softnas" {
-  source = "../../modules/user_nfs_softnas"
+  source = "../modules/user_nfs_softnas"
 
   num_instances             = 1
```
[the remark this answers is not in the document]
xoen: Oh and I also noticed that we dropped the `default_volume_size = 250`, again, apologies if it's intended.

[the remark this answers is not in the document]
kerin: Nope, that's all because `dev` and `alpha` config were slightly different, and this new version was taken from `dev` - but clearly I should've taken it from `alpha` - I'll fix.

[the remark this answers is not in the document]
xoen: 👍

infra/terraform/environments/alpha/main.tf
```diff
@@ -1,108 +0,0 @@
...
-    key    = "alpha/terraform.tfstate"
-    region = "eu-west-1"
-  }
-}
-
-provider "aws" {
-  region = "${var.region}"
-}
-
-data "aws_caller_identity" "current" {}
```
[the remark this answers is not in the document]
kerin: Not sure if that data source is still needed. The global/env resource split is pretty simple - `global` is for any resources which aren't part of an environment (parent DNS zone, S3 bucket for environment Terraform state), and environment resources are for an instance of the platform as a whole (VPC, lambda functions, IAM stuff, etc).

There's no doubt a lot of refactoring we could be doing - but not as part of this PR.

xoen:
What happened to this 😇?

(I may need to add a couple of things in terraform at some point, e.g. an RDS instance for Airflow and possibly an EFS instance. I'm wary of adding more conflicts/work for you)

PR #144, opened 2018-05-31
Prune EBS Snapshots

djsd123:
## What
Lambda function that deletes snapshots of a specified age. Optionally filter and
target snapshots by tag.

Frequency: daily

## How to review

Already deployed

1. `export DAYS_OLD=5`
2. `go run main.go`

__and__

1. mv infra/terraform/global/assets/prune_ebs_snapshots/main infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots
2. `terraform workspace new test`
3. `terraform plan -target=module.kubernetes_prune_ebs_snapshots -var-file=infra/terraform/global/assets/prune_ebs_snapshots/vars_prune_ebs_snapshots.tf` __Done__
4. `terraform apply -target=module.kubernetes_prune_ebs_snapshots -var-file=infra/terraform/global/assets/prune_ebs_snapshots/vars_prune_ebs_snapshots.tf`


infra/terraform/global/assets/prune_ebs_snapshots/README.md
```diff
@@ -0,0 +1,48 @@
...
+You'll also need a working [Go](www.golang.org/doc/) environment to build a distribution
+
+For working with the [Go](www.golang.org) runtime in AWS Lambda. See [lambda-go-how-to-create-deployment-package](https://docs.aws.amazon.com/lambda/latest/dg/lambda-go-how-to-create-deployment-package.html)
+
+To Build and Deploy
+---------
+
+The Lambda execution environment uses a Linux kernel, so you'll need to build the binary for linux
+```
+GOOS=linux go build main.go
```
xoen: You may want to add the `-o prune_ebs_snapshots` so you don't need to rename/`mv` the file later.

djsd123: You're right.  Better yet! I'll just rename the file.  As long as the package is `main`. ta

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
```diff
@@ -0,0 +1,56 @@
...
+	getDays             = os.Getenv("DAYS_OLD")
+)
+
+func main() {
+
+	// Convert `DAYS_OLD` value to an integer
+	daysOldValue, err := strconv.Atoi(getDays)
+	if err != nil {
+		fmt.Println(err)
+		log.Panicf("The type of the value you provided is %T. %v needs to be an integer", daysOldValue, daysOldValue)
```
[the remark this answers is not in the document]
xoen: Yeah: https://play.golang.org/p/g-9_mjWSYJF

So I'd probably augment the existing message, something like:

```go
// Convert `DAYS_OLD` value to an integer
daysOldValue, err := strconv.Atoi(getDays)
if err != nil {
    log.Panicf("Failed to parse DAYS_OLD, must be an integer: %s", err)
}
```

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+					Name: aws.String("owner-id"),
+					Values: []*string{
+						aws.String(ownerID),
+					},
+				},
+			},
+		}
+	}
+
+	output, err = connection.DescribeSnapshots(snapShotInput)
```
xoen: You can just return this ;)

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+	var snapShotInput *ec2.DescribeSnapshotsInput
+
+	// Get the owner ID a.k.a the aws account number
+	callerResult, callerErr := caller.GetCaller()
+	if callerErr != nil {
+		log.Panic(callerErr)
+	}
+	ownerID := *callerResult.Account
+
+	if len(key) != 0 && len(val) != 0 {
```
xoen: I can see both branches of the `if` have the filter by `owner-id`, you could have the common part outside the `if` and have a positive `if` to add the `key`/`value` if necessary (you can therefor remove the variable declaration on top of the function, e.g. (not tested ⚠️ !)

```go
func GetSnapShots(connection *ec2.EC2, key string, val string) (output *ec2.DescribeSnapshotsOutput, err error) {
	// Get the owner ID a.k.a the aws account number
	caller, err := caller.GetCaller()
	if err != nil {
		log.Panicf("Error while getting owner ID: ", err)
	}
	ownerID := *caller.Account

	filters := []*ec2.Filter{
		{
			Name: aws.String("owner-id"),
			Values: []*string{
				aws.String(ownerID),
			},
		},
	}

	if len(key) > 0 && len(value) > 0 {
		filters = append(filters, &ec2.Filter{
			Name: aws.String(fmt.Sprintf("tag:%s", key)),
			Values: []*string{
				aws.String(val),
			},
		})
	}

	snapShotInput := &ec2.DescribeSnapshotsInput{Filters: filters}

	return connection.DescribeSnapshots(snapShotInput)
}
```

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+	return output, err
+}
+
+func PruneSnapShots(connection *ec2.EC2, snapshotID string) (output *ec2.DeleteSnapshotOutput, err error) {
+
+	deleteSnapShotShotInput := &ec2.DeleteSnapshotInput{
+		SnapshotId: aws.String(snapshotID),
+	}
+
+	output, err = connection.DeleteSnapshot(deleteSnapShotShotInput)
```
xoen: As mentioned, you could just return the results from `DeleteSnapshot()` directly.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return output, err
+}
+
+func CheckSnapShotAge(SnapShotStartTimeStamp *time.Time, duration time.Duration) (isOlder bool, err error) {
+
+	isOlder = SnapShotStartTimeStamp.Add(duration).Before(time.Now())
+
+	return isOlder, err
```
xoen: As far as I can tell there is no case in which `err` wouldn't be `nil` in this function right? I think you can simply return a `bool` in this simple case.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+	deleteSnapShotShotInput := &ec2.DeleteSnapshotInput{
+		SnapshotId: aws.String(snapshotID),
+	}
+
+	output, err = connection.DeleteSnapshot(deleteSnapShotShotInput)
+
+	return output, err
+}
+
+func CheckSnapShotAge(SnapShotStartTimeStamp *time.Time, duration time.Duration) (isOlder bool, err error) {
```
xoen: You could rename few things here if you feel like it, e.g.
- This function could be renamed to `OlderThan()` as "snapshot" is implicit as you're in the `snapshots` package
- `SnapShotStartTimeStamp` => `createdAt` or something like that

Probably this could be:

```go
func OlderThan(createdAt *time.Time, duration time.Duration) bool {
    return createdAt.Add(duration).Before(time.Now())
}
```

Not a massive difference really, just a bit easier on the eyes maybe. 
Also, you may argue that this function is not really snapshots-specific so it may live somewhere else. I'd not worry too much about that tho, just something to keep in mind in case in the future you see the opportunity to factor this out.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
xoen: Sorry, it should be: `log.Panicf("Error while getting owner ID: %s", err)`

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
```diff
@@ -0,0 +1,56 @@
...
+
+	for _, snapShot := range snapShots.Snapshots {
+
+		snapShotID := *snapShot.SnapshotId
+
+		// Cast integer to type Duration
+		numberOfDays := time.Duration(daysOldValue)
+		days := time.Hour * 24 * numberOfDays
+
+		isTwoWeeksOrOlder, err := snapshot.CheckSnapShotAge(snapShot.StartTime, days)
```
xoen: As far as I can tell (see other comments), there is no way `err` is not `nil` in this utility function. Therefore you can simplify this and get rid of the `if` here 😃

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
```diff
@@ -0,0 +1,56 @@
...
+		if err != nil {
+			log.Printf("Error while evaluating the snapshot's age %s: %s", snapShotID, err.Error())
+		}
+
+		if isTwoWeeksOrOlder != false {
```
xoen: Again, see other comments but this could become something like:

```go
if snapshots.OlderThan(snapShot.StartTime, days) {
    // Prune this snapshot
}
```

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
```diff
@@ -0,0 +1,56 @@
...
+			_, err := snapshot.PruneSnapShots(connection, snapShotID)
+			if err != nil {
+				log.Printf("Error while pruning snapshot %s: %s", snapShotID, err.Error())
```
xoen: Cool thing is you don't need to explicitly convert `err` to a `string`, `Printf()` is "clever" enough, e.g.

```go
log.Printf("Error while pruning snapshot %s: %s", snapShotID, err)
```

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	snapShots, err := snapshot.GetSnapShots(connection, snapShotKeyFilter, snapShotValueFilter)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	for _, snapShot := range snapShots.Snapshots {
+
+		snapShotID := *snapShot.SnapshotId
```
xoen: Just out of curiosity, do you need to use the `*` here? Go sometimes is able to [spare you the pain](https://play.golang.org/p/lHuZHX_N0EJ) of doing it directly.

I think you already do it later when you just call `snapShot.StartTime` (without the `*`).

Where I'm getting at is that I'd probably not have this `snapShotID` variable at all and use `snapShot.SnapshotId` directly as it doesn't impact readably in fact, it's probably easier to read by removing a couple of lines of code. IMHO of course.

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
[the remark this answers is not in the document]
djsd123: Yep, spot on.  Thank G for the playground

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
djsd123: Do you mean `return output, err = connection.DescribeSnapshots(snapShotInput)`?

The compiler did not like that

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
djsd123: Thanks for this.  Did not know it was possible.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
djsd123: I figured it out

xoen: I meant `return connection.Describe...`, did it work? I thought it should.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
djsd123: 👍

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
xoen: No problem :)

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
djsd123: The function arg is type string whereas `snapShot.SnapshotId` is `*string` so what you mean is handle it like I did the start time and change the function sig?

djsd123: Like this?
```
func PruneSnapShots(connection *ec2.EC2, snapshotID *string) (output *ec2.DeleteSnapshotOutput, err error) {

	deleteSnapShotShotInput := &ec2.DeleteSnapshotInput{
		SnapshotId: aws.String(*snapshotID),
		DryRun:     aws.Bool(true),
	}

	return connection.DeleteSnapshot(deleteSnapShotShotInput)
}
```

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+
+func PruneSnapShots(connection *ec2.EC2, snapshotID string) (output *ec2.DeleteSnapshotOutput, err error) {
+
+	deleteSnapShotShotInput := &ec2.DeleteSnapshotInput{
+		SnapshotId: aws.String(snapshotID),
+	}
+
+	output, err = connection.DeleteSnapshot(deleteSnapShotShotInput)
+
+	return output, err
```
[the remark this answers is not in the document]
andyhd: I think @xoen pointed this out already

infra/terraform/global/assets/prune_ebs_snapshots/prune_ebs_snapshots.go
xoen: I see. Nah, I think it's right that `PruneSnapshots()` takes a "normal" string.
In that case leave the client code as it is, it's probably cleaner to have the variable and it's not a big deal anyway so don't worry.

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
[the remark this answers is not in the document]
xoen: Yes, I think it's possible and @djsd123 was in the process of replacing these returns 👍

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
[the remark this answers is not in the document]
xoen: In this case it's just assignment (`=`) because these variables are declared in the function signature (as they're named)

infra/terraform/global/assets/prune_ebs_snapshots/snapshot/snapshot.go
```diff
@@ -0,0 +1,59 @@
...
+			},
+		})
+	}
+
+	snapShotInput := &ec2.DescribeSnapshotsInput{Filters: filters}
+
+	return connection.DescribeSnapshots(snapShotInput)
+}
+
+func PruneSnapShots(connection *ec2.EC2, snapshotID string) (output *ec2.DeleteSnapshotOutput, err error) {
```
xoen: Just noticed that this function actually delete only one single snapshot, so I'd rename this function to `PruneSnapshot()` or even `DeleteSnapshot()` to avoid any confusion.

xoen:
@andyhd FYI: The thing of adding context to lower level errors is idiomatic of Go - You'll see that all over the place. I can't find the article where this was explained but you'll see it also in standard libraries, etc...here an example in part of the [documentation](https://tour.golang.org/methods/19).

The idea is that instead of throwing/logging an obscure low-level error (`IOerr 42` what?) you add some context about the higher level operation being performed when the error occurred. I personally found this very useful in the past.

infra/terraform/global/assets/prune_ebs_snapshots/caller/caller.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
+)
+
+func GetCaller() (output *sts.GetCallerIdentityOutput, err error) {
+
+	connection := sts.New(session.Must(session.NewSession()))
+	callerInput := &sts.GetCallerIdentityInput{}
+	output, err = connection.GetCallerIdentity(callerInput)
```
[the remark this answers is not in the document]
andyhd: Ok, I see now

djsd123:
@xoen, I changed the name of the `PruneSnapShots` function to `DeleteSnashot` as suggested.

PR #148, opened 2018-06-29
Added Airflow EFS volume and RDS DB instance

xoen:
### What
Applying this will create the EFS volume and the Postgres DB used by Airflow.

### Modules
I initially added a single module for all the Airflow resources but
I then thought that the problem of provisioning an "EFS volume" or
a "Postgres DB" are generic and common enough to deserve their own
terraform modules.

I added the `efs_volume` module which creates an EFS filesystem in the
given VPC and grant access from given security group (the "extra"
security group to which all k8s nodes belong to).

I also added the `postgres_db` module which creates a Postgres database
and allows connections from given security group (again, the k8s' "extra"
security group in our case).

### Ticket
https://trello.com/c/mM2ooFhD/1100-etl-batch-jobs-focus-on-airflow


infra/terraform/environments/dev/main.tf
```diff
@@ -7,7 +7,8 @@ terraform {
 }
 
 provider "aws" {
-  region = "${var.region}"
+  region  = "${var.region}"
+  version = ">= v1.25.0"
```
xoen: Bumped AWS provider version to use [`aws_efs_file_system.dns_name`](https://www.terraform.io/docs/providers/aws/d/efs_file_system.html) attribute.

xoen:
Thanks @kerin 👍

PR #149, opened 2018-06-29
Elasticsearch: Backup

djsd123:
Sets up AWS resources for an elasticsearch repository with an s3
backend.

### DIFF
--- a/infra/elasticsearch/iam.tf
+++ b/infra/elasticsearch/iam.tf
@@ -1,0 +1,43 @@
+data "aws_iam_policy_document" "analytics_logging_role_policy" {
+  statement {
+    actions = [
+      "s3:ListBucket",
+      "s3:GetBucketLocation",
+      "s3:ListBucketMultipartUploads",
+      "s3:ListBucketVersions",
+    ]
+
+    effect = "Allow"
+
+    resources = ["${aws_s3_bucket.analytics_logging_bucket.arn}"]
+  }
+
+  statement {
+    actions = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:DeleteObject",
+      "s3:AbortMultipartUpload",
+      "s3:ListMultipartUploadParts",
+    ]
+
+    effect = "Allow"
+
+    resources = ["${aws_s3_bucket.analytics_logging_bucket.arn}/*"]
+  }
+}
+
+resource "aws_iam_policy" "analytics_logging_policy" {
+  policy = "${data.aws_iam_policy_document.analytics_logging_role_policy.json}"
+  name   = "${var.name}"
+}
+
+resource "aws_iam_user_policy_attachment" "analytics_logging_user_policy_attachment" {
+  user       = "${aws_iam_user.analytics_logging_user.name}"
+  policy_arn = "${aws_iam_policy.analytics_logging_policy.arn}"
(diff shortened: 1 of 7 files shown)


infra/elasticsearch/iam.tf
```diff
@@ -0,0 +1,43 @@
+data "aws_iam_policy_document" "analytics_logging_role_policy" {
+  "statement" {
```
xoen: Are the quotes (`"`) alright here? I didn't realise you could use quotes here. As long as it works it doesn't matter really.

infra/elasticsearch/iam.tf
```diff
@@ -0,0 +1,43 @@
...
+    effect = "Allow"
+
+    resources = ["${aws_s3_bucket.analytics_logging_bucket.arn}"]
+  }
+
+  statement {
+    actions = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:DeleteObject",
```
xoen: Does this logging user needs to delete S3 objects?

infra/elasticsearch/variables.tf
```diff
@@ -0,0 +1,25 @@
+terraform {
+  required_version = "~> 0.11.0"
+
+  backend "s3" {
```
xoen: Stupid question: Any reason why we're keeping this in a separate state file, separate from the rest of terraform stuff? 

Also, don't know if these files should go under `infra/terraform` (they're terraform files after all). But maybe you put them here to avoid confusion by mixing with the rest?

infra/elasticsearch/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+    bucket     = "terraform.analytics.justice.gov.uk"
+    key        = "elasticsearch/terraform.tfstate"
+    region     = "eu-west-1"
+    encrypt    = true
+    kms_key_id = "arn:aws:kms:eu-west-1:593291632749:key/df8888e3-4080-4c2b-a71e-1425e72f98e4"
```
xoen: Where is this coming from?

infra/elasticsearch/variables.tf
djsd123: I don't consider this part of the cluster in that it's purely a store for a backup in order to perform an upgrade.  I just also like separation rather than the terralith model.  If it were a scheduled backup, I'd consider it a cluster operation.

infra/elasticsearch/iam.tf
djsd123: good catch although it does work it.

infra/elasticsearch/variables.tf
djsd123: KMS

infra/elasticsearch/iam.tf
djsd123: Yes, I want this user to be able to perform delete object on this bucket.  This is the user one of us is running rest calls as on the __kopf__ console.  Think of s3 as a backend rather than a datastore.

infra/elasticsearch/variables.tf
xoen: I should have been more specific 🤣 - what I meant was, "who/what created this key?"
Maybe I missed the relevant terraform bits.

PR #150, opened 2018-07-04
Airflow alpha resources (Postgres and EFS volume)

xoen:
## What
Using the modules already created in other [PR for `dev` environment](https://github.com/ministryofjustice/analytics-platform-ops/pull/148) to create DB and EFS volume used by Airflow (DB credentials are different)


xoen:
@kerin I cherry-picked your workaround to add `num_subnets` to the `efs_volume` module instead of using `length()` https://github.com/ministryofjustice/analytics-platform-ops/pull/150/commits/c3f1284b273747a2ff23c9f2929be76f88e2ecba

xoen:
Cheers @djsd123 - Applying in `alpha` now...unless there are problems (don't think so) I'll then merge.

xoen:
FYI: Only problem was that the RDS password I chose contained a slash (`/`). I fixed this and `terraform apply`-ed and it seems all good 👍

PR #152, opened 2018-07-06
Invalid Mapping

djsd123:
These templates are using an invalid type to map data to.

Supported types are strings,floats,booleans,object,etc....

__Object__ seemed the best fit as it catches all.

This has been deployed


djsd123:
@kerin It was, yes.  It's an invalid mapping type in elasticsearch 6.  I logged a ticket and elastic said it won't fly in version 6. https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html

PR #154, opened 2018-07-10
Helm: Upgrade Releases Script

djsd123:
Upgrades helm tool releases specified with the `-a` flag to their latest chart
version.

Deployed in alpha

### DIFF
--- a/scripts/upgrade_tools_releases.sh
+++ b/scripts/upgrade_tools_releases.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+set -uo pipefail
+
+APP=
+DEBUG=false
+DRY_RUN=true
+CONFIG_PATH=
+HELM_VER=2.9.1
+MOJANALYTICS_REPO=http://moj-analytics-helm-repo.s3-website-eu-west-1.amazonaws.com
+PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
+SAVE_VALUES=false
+
+usage() {
+  cat << EOF
+
+  Usage:
+
+  REQUIRED
+  -a - The Name of the application/helm chart you want to upgrade
+
+  [Optional]
+  -d - Debug: true or false, Default: false
+  -f - File: Path to config/values file, i.e. analytics-platform-config/chart-env-config/<ENV>/rstudio.yml
+  -s - Save release values yaml files: true or false, Default: false
+  -h - Print usage
+
+  Example: Upgrade all rstudio releases in dev with debug on and save the resulting values yaml file(s)
+
+    upgrade_tools_releases.sh -a rstudio -f analytics-platform-config/chart-env-config/dev/rstudio.yml -s true -d true
+
+EOF
+}
+
+while getopts a:d:f:s:h option; do
+  case $option in
+    a   ) APP="$OPTARG";;
+    d   ) DEBUG="$OPTARG";;
+    f   ) CONFIG_PATH="$OPTARG";;
+    s   ) SAVE_VALUES="$OPTARG";;
+    h   ) usage; exit 0;;
+    \?  ) usage; exit 1;;
+  esac
+done
+
+# Exit if $APP has not been set
+if [ -z $APP ]; then
+  usage; exit 1
+fi
+
+# Determine the helm chart from the argument to the -a flag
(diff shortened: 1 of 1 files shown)


scripts/upgrade_tools_releases.sh
```diff
@@ -0,0 +1,101 @@
...
+  -h - Print usage
+
+  Example: Upgrade all rstudio releases in dev with debug on and save the resulting values yaml file(s)
+
+    upgrade_tools_releases.sh -a rstudio -f analytics-platform-config/chart-env-config/dev/rstudio.yml -e dev -s true -d true
+
+EOF
+}
+
+while getopts a:e:d:f:n:s:h option; do
```
xoen: What is `n`?

scripts/upgrade_tools_releases.sh
```diff
@@ -0,0 +1,101 @@
...
+# Exit if there's no argument to the -a flag
+    if [ -z $APP ]; then
+      usage; exit 1
+    fi
+
+# Determine the helm chart from the argument to the -a flag
+HELM_REPO=mojanalytics/$APP
+
+# If CONFIG_PATH has a value prefix it with "--values"
+    if [[ ! -z $CONFIG_PATH && -f $CONFIG_PATH ]]; then
```
xoen: I think indentation is a bit off here and in few other places.

scripts/upgrade_tools_releases.sh
```diff
@@ -0,0 +1,101 @@
...
+	read -p "Is this a DRY RUN? i.e. NO-OP [y/N]: " dry_run
+	case $dry_run in
+	    [y]*    ) DRY_RUN=true; break;;
+	    [N]*    ) DRY_RUN=false; break;;
+	    *       ) echo "Answer y to simulate the operation.  Answer N to perform the operation."; exit;;
+    esac
+done
+
+# Grab the Helm client if not installed
+which helm || wget --progress=dot:giga -O helm https://storage.googleapis.com/kubernetes-helm/helm-v$HELM_VER-$PLATFORM-amd64.tar.gz
```
xoen: ~Not trolling here. I imagine this would not work on non-Macs, e.g. on Ravi's machine right?~

Actually, `$PLATFORM` is set to `uname -s` 👍

scripts/upgrade_tools_releases.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Ensure we've added our helm repo
+helm repo list | grep -F 'mojanalytics' && helm repo update || helm repo add mojanalytics $MOJANALYTICS_REPO
+
+# Get a list of usernames
+USERNAMES=$(helm ls ${APP} --max 10000 | awk '{print $10}' | sed 's/user-//g')
+
+for user in $USERNAMES; do
+
+  RELEASE_NAME=$user-$APP
```
xoen: Release names are not cut right? It's some of the deployment names which are truncated right?

scripts/upgrade_tools_releases.sh
```diff
@@ -0,0 +1,101 @@
...
+
+  helm get values $RELEASE_NAME > ${user}.yaml \
```
xoen: I think it may be worth adding a comment to explain why we're doing this `helm get values`/`helm --values`, and why we're not using `--reuse-values`, what `helm get values` returns, and probably a link to that famous helm issue/comment.

You know, for the future us/for people not familiar with helm and its gotchas.

scripts/upgrade_tools_releases.sh
djsd123: 👍

scripts/upgrade_tools_releases.sh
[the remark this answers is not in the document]
djsd123: `-f  $CONFIG_PATH` true if `$CONFIG_PATH` zero length I think!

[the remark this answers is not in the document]
andyhd: Not for me, at least:
```shell
[[ -f "" ]] && echo $?
```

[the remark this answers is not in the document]
djsd123: Oh, I see. I have`$CONFIG_PATH= ` not `""`

PR #156, opened 2018-07-23
(S) Airflow's RDS instances with more memory

xoen:
What
====
Airflow DB instance types were `t2.micro` (512 MiB of memory).

Now using `db.m3.medium` which has 3.75 GiB (on both `dev` and `alpha`).

See: https://aws.amazon.com/rds/instance-types/

Why
===
Even with relatively trivial DAGs with 30-60 tasks running concurrently
we were seeing some errors relative to DB connections.

It seemed like the DB dropped connections after ~32.
The `max_connections` depends on the amount of memory the RDS instance has,
specifically:

max_connections = DBInstanceClassMemory / 12582880

So 3.75GiB will give us a `max_connections` of ~320.

See: https://serverfault.com/a/862418

### DIFF
--- a/infra/terraform/environments/alpha/main.tf
+++ b/infra/terraform/environments/alpha/main.tf
@@ -121,10 +121,11 @@ module "airflow_storage_efs_volume" {
 module "airflow_db" {
   source = "../../modules/postgres_db"
 
-  instance_name = "${var.env}-airflow"
-  db_name       = "airflow"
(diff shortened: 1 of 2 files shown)


xoen:
FYI: `alpha` DB instance is pending modification and will change during next maintenance window (`sun:06:00-sun:08:00 UTC (GMT)`).

PR #157, opened 2018-07-30
Concourse list-roles user

r4vi:
## What

Create a new aws user that has the permission to list roles.

This will be used by concourse, specifically the airflow dags pipeline as it
will check that any roles annotated in a DAG also exist in AWS.

We may later provide a mechanism to automatically create roles based on source
code committed alongside the DAG but for now it's just a check

## How to review

1. `terraform plan`
2. see `4 added, 0 destroyed` in the output
3. `terraform apply`

https://trello.com/c/hmFr3yb8/1270-ci-for-airflow-dags

### DIFF
--- a/infra/terraform/global/main.tf
+++ b/infra/terraform/global/main.tf
@@ -66,6 +66,12 @@ module "hmpps_oasys_upload_user" {
   system_name       = "oasys"
 }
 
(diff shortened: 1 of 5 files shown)


r4vi:
I have already `apply`'d it

PR #159, opened 2018-08-09
Improve Terraform, kops, etc following manual testing [WIP]

davidread:
Notes:

* We store Terraform state in an S3 bucket, which needs creating before Terraform can run at all. This was set-up by `global.s3.terraform_state`, which can't work from scratch. So I removed that and added awscli instructions to create this bucket initially.

* Terraform commands need a `-backend-config` option for when it's not MoJ AP, since S3 bucket names need to be unique globally.

* `infra/kops/example_cluster` doesn't exist any more, and would probably get out of date anyway. Better to get users to copy the alpha one and I've written some yq foo to fill in the correct values.

* Helm RBAC setup - is this is the right way to do this? I don't think it does the same as the rbac chart, but I'd appreciate this checking.

* I added the set-up for Auth0 tenant and initial application. We should convert this to terraform this at some point.

* I renamed `xyz_...` to `platform_...`

* The gitcrypted terraform.tfvars diff is:
```
-terraform_bucket_name = "..."
-terraform_base_state_file = "..."
-xyz_root_domain = "..."
+platform_root_domain = "..."
```

* softnas inc stopped allowing new customers on the existing AMI (`ami-6498ac02`, which remains the default), and in these instructions, new platforms are directed to a new one (`ami-22cecec8`).

TODO:
* [x] Run terraform against dev/alpha to see if it is up to date.
* [ ] Get rid of infra/terraform/global/godobject.tfvars
* [x] Check Helm RBAC setup is right
* [ ] Uncomment `module "data_backup"`


davidread:
I've split this into smaller tickets, apart from this straggler:
```
/infra/terraform/platform/main.tf
-module "data_backup" {
-  source = "../modules/data_backup"
-
-  env                 = "${terraform.workspace}"
-  k8s_worker_role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/nodes.${terraform.workspace}.${data.terraform_remote_state.base.xyz_root_domain}"
-  logs_bucket_arn     = "${data.terraform_remote_state.base.s3_logs_bucket_name}"
-}
+# module "data_backup" {
+#   source = "../modules/data_backup"
+#
+#   env                 = "${terraform.workspace}"
+#   k8s_worker_role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/nodes.${terraform.workspace}.${data.terraform_remote_state.base.platform_root_domain}"
+#   logs_bucket_arn     = "${data.terraform_remote_state.base.s3_logs_bucket_name}"
+# }
```

davidread:
Closing as the bulk of this work is split into other tickets and all merged now.

PR #160, opened 2018-08-14
Cert-Manager: EC2 Role

djsd123:
[Trello](https://trello.com/c/M1snktNZ)

Relates to [PR](https://github.com/ministryofjustice/analytics-platform-helm-charts/pull/205)

Module to create instance role for [DNS01](https://cert-manager.readthedocs.io/en/latest/tutorials/acme/dns-validation.html) challenge

Required access for this process is detailed [here](https://cert-manager.readthedocs.io/en/latest/reference/issuers/acme/dns01.html#amazon-route53)


infra/terraform/platform/main.tf
```diff
@@ -119,3 +119,10 @@ module "airflow_db" {
   node_security_group_id = "${module.aws_vpc.extra_node_sg_id}"
   subnet_ids             = "${module.aws_vpc.storage_subnet_ids}"
 }
+
+module "cert_manager" {
+  source           = "../modules/ec2_cert_manager_role"
+  role_name        = "${var.role_name}"
```
xoen: Any reason why the role name doesn't start with the environment name as usual?

Also, it's probably not sensitive? You could potentially use string concatenation in the `main.tf` instead of having it encrypted.

djsd123: Do you mean `alpha-cert-manager` instead of `cert-manager-alpha`?

Example of concat?

xoen: Yeah, couldn't you use something like this here?

```
role_name        = "${terraform.workspace}-cert-manager"
```

My TF is rusty so maybe it's not possible for some reason.

xoen: BTW, not a problem, just a thought as in theory that role is not sensitive and we can be more explicit/clear by having it inline instead of in a separated, encrypted, file.

But feel free to ignore.

djsd123: Oh right, yes I see what you mean.  Sure I'll change it

PR #162, opened 2018-09-07
Imported SES resources for root domain

xoen:
The SES resources were created manually in the AWS Console rather than using Terraform.
This PR imports these resources so that they're managed by Terraform.

### DIFF
--- a/infra/terraform/modules/ses_domain/main.tf
+++ b/infra/terraform/modules/ses_domain/main.tf
@@ -1,3 +1,37 @@
-resource "aws_ses_domain_identity" "root_domain" {
-  domain = "mojanalytics.xyz"
+resource "aws_ses_domain_identity" "domain" {
+  domain = "${var.domain}"
+}
+
+# SES Verification: TXT Record
+resource "aws_route53_record" "amazonses_verification_record" {
+  zone_id = "${var.aws_route53_zone_id}"
+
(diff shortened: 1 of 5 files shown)


infra/terraform/modules/ses_domain/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  depends_on = ["aws_route53_record.amazonses_verification_record"]
+}
+
+# SES Verification: DKIM
+resource "aws_ses_domain_dkim" "domain_verification" {
+  domain = "${aws_ses_domain_identity.domain.domain}"
+}
+
+resource "aws_route53_record" "domain_amazonses_dkim_verification_record" {
+  count   = 3
```
kerin: I'd say this should be `count = ${length(aws_ses_domain_dkim.domain_verification.dkim_tokens)}`, in case AWS change the number of DNS records used for DKIM verification.

xoen: Sorted: https://github.com/ministryofjustice/analytics-platform-ops/pull/162/commits/db400cc48f530a7ef9607f923f14a9925dc8c8d1

xoen:
@kerin FYI: I force-pushed to get rid of the non-related ingress commit. Rest is the same.

PR #163, opened 2018-09-10
Added SES SMTP user for Airflow

xoen:
The `ses_smtp_user` can be used to create an IAM user with
permission to send emails from the specified SES Domain Identity.

**NOTE**: The SMTP password for the user is not specified but it's
an output of the IAM Access Key associated with the IAM user

Luckily terraform's [`iam_access_key` has an attribute to get hold of the SMTP password](https://www.terraform.io/docs/providers/aws/r/iam_access_key.html#ses_smtp_password) without having to do anything weird.

Ticket: https://trello.com/c/tTZ1DG3y/1322-fix-airflow-email-notifications


xoen:
Thanks @kerin :)!

PR #164, opened 2018-09-12
Reworked `ses_smtp_user` module as SMTP username is IAM Access Key ID

xoen:
### SMTP username needs to be IAM Access Key ID
I thought it would be the IAM user name. I was wrong.
As the SMTP password is generated from the IAM Access Key Secret,
it makes sense for the SMTP username to be the IAM Access Key ID.
This means the SMTP user is really the IAM Access Key which can then
be revoked if necessary.

**NOTE**: As the SMTP password can only be used to send email if
compromised and even if the user has other IAM permissions this will
not be a disaster (that's why by design the SMTP password is not just
the IAM Access Key Secret). Not that this user should really have
more permissions than just the ability to send emails (Principle
of Least Privilege)

###  IAM Policy change
In the IAM policy the `Resource` needs to be the SES email address identity ARN, not the SES domain identity ARN. This wasn't very clear from the documentation I've read but needs to change as well.

### Ticket
https://trello.com/c/tTZ1DG3y/1322-fix-airflow-email-notifications


xoen:
@r4vi let me know if anything is not clear

r4vi:
so we have to create the identity manually in the console and pass ARN in to terraform?

xoen:
@r4vi correct. I didn't find a Terraform resource for email identity and they would need manual verification anyway. 

Thanks for the review 👍

PR #165, opened 2018-09-14
Update kops instancegroups to match deployed spec

kerin:
Node numbers and master instance types had been edited in place in the alpha cluster - this change gets git and the cluster back in sync


davidread:
Stale

PR #171, opened 2018-09-27
Idled Ingress Class fixer

r4vi:
this script fixes ingresses in a bad state for idled rstudio instances.
If the ingress object for the user is still enabled but they are idled then they
can never be un-idled because the request will try to reach their pod instead of
the unidler.
This script loops through all idled deployments and then gets the corresponding
ingress object, the object is then checked for having an nginx ingress class
annotation. If it has this annotation then it is changed to 'disabled'

### DIFF
--- a/scripts/fix_idled_ingress_class.sh
+++ b/scripts/fix_idled_ingress_class.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+###
+# Find any deployments that are idled and set their ingress to 'disabled'
+###
+
+while read obj ns
+do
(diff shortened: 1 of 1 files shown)


davidread:
That's neat!

scripts/fix_idled_ingress_class.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+###
+# Find any deployments that are idled and set their ingress to 'disabled'
+###
+
+while read obj ns
```
xoen: 😱

PR #173, opened 2018-09-27
Fix circular terraform state bucket creation problem

davidread:
## What

We store Terraform state in an S3 bucket, which needs creating before Terraform can run at all. This was set-up by global.s3.terraform_state, which can't work from scratch. So I removed that and added awscli instructions to create this bucket initially.

Terraform commands need a -backend-config option for when it's not MoJ AP, since S3 bucket names need to be unique globally.

## How to review

Run terraform global and terraform platform with a fresh bucket name, and see no errors.


xoen:
Only thing worth checking with @kerin is whether removing this resource in the terraform files would confuse TF and cause it to try to delete this S3 bucket (maybe this was the reason why this was in terraform, so that he's aware of it). But I'm not really sure.

davidread:
@xoen yep, good catch. I've just run this to edit the terraform state, so the terraform state S3 bucket won't get deleted:

     cd infra/terraform/global
     terraform state rm aws_s3_bucket.terraform_stateterraform_state

davidread:
@xoen yep, good catch. I've just run this to edit the terraform state, so the terraform state S3 bucket won't get deleted:

     cd infra/terraform/global
     terraform state rm aws_s3_bucket.terraform_stateterraform_state

PR #174, opened 2018-09-27
automate kops config

davidread:
## What

`infra/kops/example_cluster` doesn't exist any more, and would probably get out of date anyway. Better to get users to copy the alpha one and I've written some yq foo to fill in the correct values. It should probably be in a separate script, but maybe this is ok for now.

## How to review

Create a cluster, see there are no errors.


davidread:
Check fails only because softnas variables are introduced on PR #177

README.md
```diff
@@ -137,41 +137,80 @@ Once complete your base AWS resources should be in place
...
+```
+
+3. Set the correct values for your new cluster config:
+```
+cd infra/terraform/global
+export KOPS_STATE_STORE=s3://`terraform output kops_bucket_name` >/tmp/kops_bucket_name
+
+cd ../../../infra/terraform/platform
+export ENV_DOMAIN=`terraform output -module=cluster_dns dns_zone_domain`
+yq w -i ../../../infra/kops/clusters/$ENVNAME/cluster.yml spec.configBase $KOPS_STATE_STORE/$ENV_DOMAIN
```
xoen: Nice. I guess copy and paste from README is less error prone.

PR #175, opened 2018-09-27
add helm rbac setup

davidread:
## What

This is how I got helm to work, by deploying this config that was already in this repo.

## How to review

Deploy the cluster up to this point and see if helm commands work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,17 @@ Once complete your base AWS resources should be in place
 
 If kubectl is unable to connect, the cluster is still starting, so wait a few minutes and try again; Terraform also creates new DNS entries, so you may need to flush your DNS cache. Once `cluster-info` returns Kubernetes master and KubeDNS your cluster is ready.
 
+### Helm RBAC setup
+
+Helm/Tiller should use its own service account. Create it like this:
+```
+kubectl create -f config/helm/helm.yml
+# Tell helm to use it
+helm init --service-account helm
+# Check it creates
+kubectl describe deployment tiller-deploy -n kube-system -f
+```
+
 ### Modifying AWS and cluster post-creation
 Once all of the above has been carried out, both Terraform and Kops state buckets will be populated, and your local directory will be configured to push/pull from those buckets, so changes can be made without further configuration.
 
--- a/README.md
+++ b/README.md
@@ -182,13 +182,13 @@ If kubectl is unable to connect, the cluster is still starting, so wait a few mi
 
 ### Helm RBAC setup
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -180,6 +180,17 @@ Once complete your base AWS resources should be in place
 
 If kubectl is unable to connect, the cluster is still starting, so wait a few minutes and try again; Terraform also creates new DNS entries, so you may need to flush your DNS cache. Once `cluster-info` returns Kubernetes master and KubeDNS your cluster is ready.
 
+### Helm RBAC setup
+
+Helm/Tiller should use its own service account. Create it like this:
+```
+kubectl create -f config/helm/helm.yml
```
r4vi: This is kind of weird because in `alpha` we have an account called `tiller` and in dev we have one called `helm`. 
We should pick one and stick with it - `helm` is fine I guess but @djsd123 might have a reason why he used `tiller` in alpha

davidread: Ah, good spot. 'tiller' makes more sense as that's what the helm docs use, so I'll change the name to that.

README.md
```diff
@@ -180,6 +180,17 @@ Once complete your base AWS resources should be in place
 
 If kubectl is unable to connect, the cluster is still starting, so wait a few minutes and try again; Terraform also creates new DNS entries, so you may need to flush your DNS cache. Once `cluster-info` returns Kubernetes master and KubeDNS your cluster is ready.
 
+### Helm RBAC setup
+
+Helm's Tiller should use its own service account. Create it like this:
+```
+kubectl create -f config/helm/tiller.yml
+# Tell helm to use it
+helm init --service-account helm
```
xoen: This is *not* creating the `helm` service account, it's just telling tiller to use the service account just created with `kubectl create -f config/helm/tiller.yml` (above) right?

If that's the case, I think this should be `$ helm init --service-account helm` after the renaming to `tiller`.

davidread: Do you mean it should be `helm init --service-account tiller`?

xoen: I think so. If that's the service account created by `$ kubectl create -f config/helm/tiller.yml` we now changed the service account name to `tiller`, no? 

https://github.com/ministryofjustice/analytics-platform-ops/pull/175/files#diff-c5d78363155c97bdda457267956dc012R4

But I may be wrong.

davidread: I've made this change: https://github.com/ministryofjustice/analytics-platform-ops/commit/9e36b7a532f5d54228cbc683e26cc3571d862eb5

xoen: 🏆

PR #176, opened 2018-09-27
Initial auth0 setup added

davidread:
## What

I added the set-up for Auth0 tenant and initial application. We should convert this to terraform this at some point.

## How to review

Follow the instructions and check you get what is seen in dev/alpha in auth0 dashboard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,41 @@ User network home directories are provided by [SoftNAS from AWS Marketplace](htt
 
 ### Defining new environment
 
+#### Auth0
+
+1. Create a new tenant:
+
+    1. Log-in to Auth0
+    2. Click on your user
+    3. In the drop-down menu click "Create tenant"
+         * Tenant domain: include the environment name, if not the platform
+
+2. Create an application:
+
+    1. In the side-bar click "Applications"
+    2. Click "Create Application"
+         * Name: AWS
+         * Application Type: Regular Web Applications
+    3. Click "Save"
+    4. Click "Settings"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,41 @@ User network home directories are provided by [SoftNAS from AWS Marketplace](htt
...
+    1. Log-in to Auth0
+    2. Click on your user
+    3. In the drop-down menu click "Create tenant"
+         * Tenant domain: include the environment name, if not the platform
+
+2. Create an application:
+
+    1. In the side-bar click "Applications"
+    2. Click "Create Application"
+         * Name: AWS
```
xoen: Does this name needs to be `AWS Console`? You mention this in step 3.2 later on right?

README.md
```diff
@@ -79,6 +79,41 @@ User network home directories are provided by [SoftNAS from AWS Marketplace](htt
...
+         * Allowed Callback URLs: `https://signin.aws.amazon.com/saml, https://aws.services.$env.$domain/callback` (replace the $variables)
+         * Allowed Web Origins: `https://aws.services.$env.$domain` (replace the $variables)
+    5. Click "Save changes"
+
+    Record the Domain and Client ID values - you'll use them in your .tfvars file in a moment.
+
+3. Download SAML2 metadata:
+
+    1. In the side-bar click "Applications"
+    2. Click "AWS Console" (created in previous step)
```
xoen: ...or maybe this needs to be just `"AWS"` assuming you're talking about the application created in step 2.2. (but it seems like we're calling this application just "AWS" in the AP Auth0 account.

davidread: good catch! I was in two minds. I'll make all these "AWS"

PR #177, opened 2018-09-27
add softnas setup detail

davidread:
## What

Softnas inc stopped allowing new customers on the existing AMI (ami-6498ac02, which remains the default), and in these instructions, new platforms are directed to a new one (ami-22cecec8).

## How to review

Run the setup to create a softnas and see ideally use the params in a working install of the platform.


README.md
```diff
@@ -71,14 +71,47 @@ terraform plan
...
+3. Wait 30 seconds before the flash message appears "Thank you for subscribing to this product!"
+4. Click "Continue to Configuration" (which has also appeared)
+5. Configure:
+
+   * Region: choose the same as chosen for the rest of your platform (e.g. EU Ireland)
+
+6. Click "Continue to Launch"
+
+   * EC2 Instance Type - select a suitable one, considering cost. Record the instance type (e.g. `m5.large`) - you'll use this in your .tfvars file in a moment.
+   * Key pair - create one called "softnas-$ENVNAME" (replacing the $ENVNAME) and save the private key (.pem file) locally
```
xoen: Do we need to do something special about the softnas key pair? This is not kept by AWS and we'll need it to access softness right?

infra/terraform/modules/user_nfs_softnas/main.tf
```diff
@@ -14,6 +14,10 @@ variable "num_instances" {
   default = 2
 }
 
+variable "softnas_ami_id" {
```
xoen: I know it doesn't matter from Terraform (TF) perspective but I've been trying to structure TF modules to have input variables in the `inputs.tf` and output variables in the `outputs.tf`. 

I like that because then when I have a look at a module I didn't personally wrote I know that by looking at the `inputs.tf` I can have an idea of what input that module is expecting, etc...

infra/terraform/modules/user_nfs_softnas/ec2.tf
```diff
@@ -115,7 +115,7 @@ resource "aws_network_interface" "softnas_eth1" {
 }
 
 resource "aws_instance" "softnas" {
-  ami                  = "${data.aws_ami.softnas.id}"
+  ami                  = "${var.softnas_ami_id}"
```
xoen: Nice one.

infra/terraform/modules/user_nfs_softnas/main.tf
davidread: Sounds good. I've renamed this file inputs.tf.

README.md
xoen: I think this is the case David so it may be the case of adding a note in the readme in case someone throw these keypair file away thinking it can be retrieved somehow.

davidread: I've added a note about the key here: https://github.com/ministryofjustice/analytics-platform-ops/pull/191/files#diff-04c6e90faac2675aa89e2176d2eec7d8R186

PR #181, opened 2018-09-27
"xyz_" variables renamed to "platform_"

davidread:
Rename because xyz refers to our mojanalytics.xyz domain name (= 'global' terraform). Better to use a word that is understandable for other orgs. Chose 'platform'.

This PR is split off from #159.


davidread:
To avoid the dns zone being recreated by terraform, I've run this to rename a resource in the terraform backend state on S3:

     cd infra/terraform/global
     terraform state mv aws_route53_zone.xyz_zone aws_route53_zone.platform_zone

PR #182, opened 2018-09-27
Scripts for checking live invariant properties of the cluster

r4vi:
This is a simple framework for checking if things are fishy with the cluster.

Initially there is only one check. If there are any deployments marked as
"idled" that have > 0 replicas.

Examples of things we can add in future:

* user's url is in unidler but not in the tls block
* user is running old version of rstudio / jupyter
* resources left in cluster for users that have left
* pod iam annotations match iam ns annotations


r4vi:
I should add a readme on how to run

scripts/invariant_checks/.eslintrc.js
```diff
@@ -0,0 +1,20 @@
...
+    "rules": {
+      "camelcase": ["off"],
+      "no-alert": ["off"],
+      "import/no-dynamic-require": ["off"],
+      "no-use-before-define": ["error", { "classes": false }],
+      "no-param-reassign": ["error", { "props": false }],
+      "no-underscore-dangle": ["error", { "allow": ["_json"] }],
+    }
+  };
+
```
r4vi: add new line

davidread:
Looking good!

scripts/invariant_checks/README.md
```diff
@@ -0,0 +1,37 @@
+# Invariant Checker
+
+This is a simple framework for checking if things are fishy with the cluster.
```
xoen: 🐟

scripts/invariant_checks/checks/idled-deployments-with-more-than-0-replicas.js
```diff
@@ -0,0 +1,25 @@
+function isIdled(deployment) {
+  return deployment.metadata.labels && deployment.metadata.labels["mojanalytics.xyz/idled"] == "true"
```
xoen: If we're sure this is coming out as a string (and it makes sense as they're labels) I may use the `===` to be on the safe side considering how weird JS can be.

scripts/invariant_checks/checks/missing-tls-block.js
```diff
@@ -0,0 +1,52 @@
...
+        spec: {
+          tls: [{
+            hosts: [spec.rules[0].host]
+          }]
+        }
+      }
+    }
+
+    try {
+      let updated = await client.apis.extensions.v1beta1.namespace(
```
xoen: I'm not familiar with `async`/`await` but it seems quite intuitive and I *think* I get the gist of it.

So the result of patch is async (a Promise) and here we wait for each of them. Potentially we could accumulate these in an array and then loop after this loop to wait for them so potentially they would be submitted concurrently...but to be honest doesn't really matter :P and maybe this is easier to follow.

xoen:
Very nice job @r4vi 🌟

scripts/invariant_checks/checks/missing-tls-block.js
xoen: Good point. To be fair, probably these patch operations would be relatively quick anyway so maybe difference is not going to be big, at least at this stage.

As said, no need to change as parallelism is more "confusing".

PR #184, opened 2018-09-27
Add a policy that is needed by controlpanel

davidread:
## What

This policy is seen on dev and alpha and is needed.


xoen:
Nice one. Thanks @davidread and sorry it wasn't already there, probably my fault!

davidread:
I added the existing policy to our terraform state like this:
```
cd infra/terraform/platform
terraform workspace select dev
terraform import aws_iam_policy.read-user-roles-inline-policies arn:aws:iam::593291632749:policy/dev-read-user-roles-inline-policies
terraform workspace select alpha
terraform import aws_iam_policy.read-user-roles-inline-policies arn:aws:iam::593291632749:policy/alpha-read-user-roles-inline-policies
```

PR #186, opened 2018-10-02
IAM user for Councourse pipeline which updates mojanalytics helm repo

xoen:
Councourse pipeline is here: https://github.com/ministryofjustice/analytics-platform-concourse-pipelines#update-helm-repoyaml

The pipeline only needs to read and write from the S3 bucket used as
Helm repository.

This will obviously run only in an environment so I've put this IAM
user in the global terraform file.

### Ticket
https://trello.com/c/DMx96N37/1331-iam-role-to-update-helm-repo-with-concourse-terraform


infra/terraform/global/variables.tf
```diff
@@ -25,3 +25,7 @@ variable "global_cloudtrail_bucket_name" {
 variable "s3_logs_bucket_name" {
   default = "moj-analytics-s3-logs"
 }
+
+variable "helm_repo_s3_bucket_name" {
+  default = "moj-analytics-helm-repo"
```
r4vi: I wonder what created this in the first place? :thinking:

xoen:
@r4vi FYI: I've applied this, so we now have IAM user.

infra/terraform/global/variables.tf
xoen: 💥 🐔 🥚

r4vi:
@xoen I'll add to config repo

infra/terraform/global/variables.tf
xoen: I suspect ✋-lly. 

I wonder if I should actually have a module that:
1. Creates that S3 bucket
2. Creates the IAM user, etc...

I also wonder if we should keep the name of IAM user that can update that S3 bucket/helm repository CI-agnostic and not have "concourse" in it.

xoen: Fair enough. I'll leave as it is as it's working 👍

PR #187, opened 2018-10-02
Update dev ClusterSpec to match deployed config

kerin:
- changed kops channel from alpha to stable
- updated Kubernetes to v1.10.5
- add R53 IAM policy to allow Traefik to query DNS zones


r4vi:
@kerin no træfik anymore so is this still required:
"add R53 IAM policy to allow Traefik to query DNS zones"

PR #197, opened 2018-10-08
Script: add helper script to put `idleable` annotation on deployed tools

r4vi:
This script will add the idleable annotation to any existing rstudio or
jupyter-lab deployments.

Just putting here for reference but really it'll only be run once

## How to review
I've already run this on dev but... delete the annotation from a deployment and run this again, it will add the annotation back.

### DIFF
--- a/scripts/label_jupyter_and_rstudio_deployments.sh
+++ b/scripts/label_jupyter_and_rstudio_deployments.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -ex
+
+IDLEABLE_LABEL=${1:-"mojanalytics.xyz/idleable"}
+
+for user in $(kubectl get ns | grep user- | cut -f 1 -d ' ' | sed 's/^user-//'); do
+    namespace=user-$user
+
+    kubectl label deployments -l app=rstudio ${IDLEABLE_LABEL}=true -n $namespace
+    kubectl label deployments -l app=jupyter-lab ${IDLEABLE_LABEL}=true -n $namespace
+
+done


scripts/label_jupyter_and_rstudio_deployments.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -ex
+
+IDLE_LABEL=${1:-"mojanalytics.xyz/idleable"}
+
+for user in $(kubectl get ns | grep user- | cut -f 1 -d ' ' | sed 's/^user-//'); do
```
xoen: Maybe I'm missing something but what about? 

```bash
for namespace in $(kubectl get ns | grep user- | cut -f1 -d" "); do
# ...
```
You're really interested in the namespaces right?

I mean maybe you used sed to be sure you're considering only namespaces starting with `user-`...potentially you could use `grep -e "^user-"` but maybe it's overkill?

scripts/label_jupyter_and_rstudio_deployments.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -ex
+
+IDLE_LABEL=${1:-"mojanalytics.xyz/idleable"}
```
xoen: Not that it matters really but maybe `IDLE_LABEL` => `IDLEABLE_LABEL`

scripts/label_jupyter_and_rstudio_deployments.sh
r4vi: you're right, I just copied this from `analytics-platform-ops/scripts/upgrade_rstudio_pods.sh` without thinking too much. I'm not going to change it because it works and this is a throw away script :wastebasket:

xoen: Happy with that ;)

PR #199, opened 2018-10-09
Kops oidc client correction

davidread:
* Correct the oidc client id to be the "kubectl-oidc" one (not the "AWS" client that terraform knows about)
* Moves kops configurer into a shell script - more manageable when there is an error
* Deletes a stray pipe


r4vi:
I think the formatting of the readme is broken somehow (missing `?)

PR #201, opened 2018-10-09
Invariant Check: Ghosted by the idler

r4vi:
Add a check for deployments the idler disabled but didn't add to the unidler
ingress, this can happens if the idler crashes half way through.

It also takes a --fix flag to automatically fix the unidler ingress in the case
of any failures.

1. Idle yourself or find an idled deployment
1. remove it from the unidler ingresses spec.rules host block
1. run `node index.js` - expect it to tell you this is missing
1. run `node index.js --fix` expect it to fix it
1. run `node index.js` expect it to fix it to pass
1. check the unidler ingress, see that the host block has been re-added


r4vi:
@davidread this doesn't manually unidle, it completes an incomplete idling (if idler died half way through)

davidread:
@ravi ok, got it. So the same issues but in reverse. Maybe it's overkill, but it would be nice if it fixed all potential problems with idling. Or maybe we just need to fix idler and call that from here... ;-)

PR #205, opened 2018-10-22
Add customization of login page

davidread:
I've added login.html to the dev and alpha branches of https://github.com/ministryofjustice/analytics-platform-auth0 so this explains how to deploy those.

Related: https://github.com/ministryofjustice/analytics-platform-auth0/pull/36


davidread:
Good thinking! That is a simple switch.

PR #206, opened 2018-10-22
Remove unused filtering to find the softnas ami

davidread:
The AMI is specified explicitly now, rather than using this 'filtering'. The only reference to this code (`data.aws_ami.softnas`) was removed in commit https://github.com/ministryofjustice/analytics-platform-ops/commit/08cf30206797c0c36eac37a28ee4ce023aa586c3


r4vi:
@davidread did we do this because we had some undesirable behaviour when just using the latest one from softnas?

r4vi:
oh I see, it's just a cleanup

PR #209, opened 2018-10-23
add terraform tip - weird terraform/pip/macos/homebrew thing

davidread:
This problem was experienced by me and Andy I seem to remember.


davidread:
Good point @kerin - I'd assumed it was a false warning because I've used pip plenty since, but that's always been in virtualenvs. I have another way pending that solves this better.

davidread:
Closed in favour of https://github.com/ministryofjustice/analytics-platform-ops/pull/212

PR #210, opened 2018-10-23
Certifi package with that date no longer exists. 

davidread:
## What

The certifi package specified no longer exists.

certifi is designed so that you always use the latest, so remove the pinning.

## How to review

Rerun terraform global for a fresh platform.


davidread:
@kerin @r4vi Any concerns about this going in?

r4vi:
I guess it's good to unpin it as we always wants the latest root certs

PR #211, opened 2018-10-23
switch user home dirs from SoftNAS to EFS [WIP]

davidread:
## What

We're talking about switching from Softnas to EFS. This PR shows the terraform changes, so we can discuss it / try this out.

@kerin was mentioned resurrecting user_nfs module from: https://github.com/ministryofjustice/analytics-platform-ops/pull/161/commits/fbefb16873d5631d7e83f37a084a08c45cb7e5ed#diff-9ba6f76a58c1a14d493175704a64c27f
which boils down to: https://github.com/ukgovdatascience/analytics-platform-ops/commit/2814da6173c0e5b6fa1ace5529796dd956bca0e7
However that is exactly the same as efs_volume module, so here I've just reused that.

The DNS stuff needs changing too. I'll leave that for now.


davidread:
see kerin's version of the same :) https://github.com/ministryofjustice/analytics-platform-ops/tree/efs-homes-performance-test

PR #213, opened 2018-10-30
Add instructions about Google sign-in

davidread:
## What

This reflects what we have on alpha. Without this setup, end-users get offered Google as an option when signing into e.g. Jupyter.

I guess we use Google sign-in for some admin/backend things. End users should always use github to login.


djsd123:
@davidread, Just [Kibana](http://kibana.services.alpha.mojanalytics.xyz) AFAIK.

PR #223, opened 2018-11-19
Generate subnet CIDRs from specified AZs

kerin:
Currently subnet CIDRs are specified as static lists, which allows for an incorrect number of resources to be created for the given AZ list - for example, 3 EFS mountpoints were being created when 1 AZ was specified. This PR instead uses Terraform’s `cidrsubnet` template function to calculate CIDR blocks based on the VPC CIDR and AZ list.


djsd123:
Clever.  Never used that function before 👍

PR #231, opened 2018-12-03
Concourse Database

djsd123:
[Trello](https://trello.com/c/j2Ll7zr4)

Migrating Concourse database from in cluster to RDS

- Adding endpoint output to the postgres_db module

- Adding Name tag to the postgres_db module


xoen:
did we give up on this because of potential concourse upgrade? (just curious)

djsd123:
Yep.  Then we gave up on upgrade because Concourse v4 had a bug with OIDC

PR #235, opened 2018-12-04
Rename AWS SAML provider to $ENV-auth0

kerin:
AWS SAML provider suffixes had been changed from ‘auth0’ to the more generic ‘idp’, but references in auth0 rules and user IAM role trust policies had not been updated to match. This commit reverts that change for now.


r4vi:
:cry:

PR #238, opened 2018-12-12
Replace kops cluster spec python script with Terraform equivalent

kerin:
Currently a python script is used to generate kops Cluster and InstanceGroup specs from config YAML files and terraform outputs. This is somewhat clumsy and opaque, and requires extra tooling and steps during environment creation. This PR begins work on replacting that python script with terraform templates.

# How to test

These instructions will create `platform-base` resources and a kubernetes cluster, but not all the other platform components. 

**Note**: the use of stdin input for `kops delete` below requires kops v1.11+

## Setup
Create an individual auth0 account/tenant first (this is free). This is currently required because AWS handles federated identity providers as globally unique, so each platform environment must have a unique IdP. This can likely be changed, but not as part of this PR.

```
$ export envname=yourtestenv

# Terraform bit
$ cd infra/terraform/platform-base
$ terraform workspace create $envname
$ cp vars/kerintest.tfvars vars/${envname}.tfvars
# edit your tfvars with correct values for your Auth0 tenant
$ terraform create -var-file=vars/${envname}.tfvars

# Kops bit
$ terraform output kops_spec | kops create -f -
$ kops create secret --name ${envname} sshpublickey admin -i ~/.ssh/id_rsa.pub
$ kops update cluster ${envname} --yes

# Validate
$ watch kops validate cluster
```

Once kops reports the cluster as healthy you should be able to successfully run `kubectl get pods` etc

## Teardown

```
$ terraform output kops_spec | kops delete -f - --yes

# be really sure you're in the right workspace
$ terraform workspace show
$ terraform destroy -var-file=vars/${envname} --yes
$ terraform workspace delete ${envname}
```

:party_parrot:


infra/terraform/modules/kops_spec/main.tf
```diff
@@ -0,0 +1,138 @@
+data "template_file" "etcd_member" {
```
xoen: So to state the obvious, this will render the template string/file interpolating the variables and then create/update these resources as if you wrote them yourself, right?

(Sorry for the silly question, I'm not familiar with `template_file` and documentation just says "Renders a template from a file")

xoen: Sorry, I didn't understand it at all :facepalm: . This will generate config files for kops, they'll be rendered and we can then copy them from the output variable.

kerin: Yep, `template_file` just takes some input variables, interpolates them into a template file and makes the result available via the `data.template_file.$name.rendered` attribute. As everything kops needs to know is available from the `aws_vpc`, `cluster_dns` and `federated_identity` modules in `platform_base`, all kops resources (`ClusterSpec`, `InstanceGroup`) can be generated within terraform and then piped into kops: `terraform output kops_spec | kops create -f -`

PR #244, opened 2018-12-19
Set up Auth0 for platform

andyhd:
## What

* Added a terraform module to setup an Auth0 tenant with necessary clients, connection and rules for use with the platform
  * Create a Github social connection
  * Create an AWS application
  * Create the `kubectl-oidc` application
  * Set Rules (with configuration keyvals)

This module is not at all useful beyond the MOJ Analytical Platform, so some work needs to be done to make it less tightly coupled to our own setup


djsd123:
You need to do a `terraform fmt`

infra/terraform/modules/auth0/main.tf
```diff
@@ -0,0 +1,87 @@
+provider "auth0" {
+  domain        = "${var.auth0_tenant_domain}"
+  client_id     = "${var.auth0_api_client_id}"
+  client_secret = "${var.auth0_api_client_secret}"
+}
+
+resource "auth0_client" "aws" {
+  name        = "test-AWS"
```
djsd123: Might be better to have this set later with a variable?

djsd123: Ignore ☝️

infra/terraform/modules/auth0/main.tf
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+  enabled_clients = [
+    "${auth0_client.aws.id}",
+    "${auth0_client.kubectl-oidc.id}",
+  ]
+}
+
+locals {
+  auth0_rules_config = "${map(
```
kerin: Any reason you can't use a literal map here?

```
auth0_rules_config = {
  foo = "bar"
}
```

infra/terraform/modules/auth0/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  )}"
+}
+
+resource "auth0_rule_config" "config" {
+  count = "${length(keys(local.auth0_rules_config))}"
+  key   = "${element(keys(local.auth0_rules_config), count.index)}"
+  value = "${element(values(local.auth0_rules_config), count.index)}"
+}
+
+variable "auth0_rules" {
```
kerin: I'd split all `variable` declarations out into a `variables.tf` file.

infra/terraform/modules/auth0/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "aws_client_id" {
```
kerin: I think we want all clients' OIDC client ID as outputs so they can be easily grabbed from `terraform output`, or passed into other modules like `federated_identity`.

infra/terraform/modules/auth0/rules/Multifactor-Google-Authenticator-Do-Not-Rename.js
```diff
@@ -0,0 +1,54 @@
+/*
+ * This rule is managed by Terraform, DO NOT EDIT!
+ */
+function (user, context, callback) {
+    var rangeCheck = require('range_check');
+
+    var AUTHENTICATOR_LABEL = 'MOJ Analytical Platform (Dev)';
```
kerin: Should interpolate the terraform `env` var into this label - which would mean templating this file (and probably all rules JS): https://www.terraform.io/docs/providers/template/d/file.html

infra/terraform/modules/auth0/rules/Multifactor-Google-Authenticator-Do-Not-Rename.js
```diff
@@ -0,0 +1,54 @@
...
+    var CONNECTION = user.identities[0].connection;
+    var ENABLED_CONNECTIONS = [
```
kerin: This should be a variable, with its contents taken from the declared client resources

```
var ENABLED_CONNECTIONS = [
  "${auth0_connection.github.name}"
]
```

Or similar - you'd actually need to use a `var.foo` here if this is changed to a template file, with the list of connections passed in to the `template_file` resource.

infra/terraform/modules/auth0/rules/Multifactor-Google-Authenticator-Do-Not-Rename.js
```diff
@@ -0,0 +1,54 @@
...
+        'github',
+        'google-oauth2',
+    ];
+    var CORPORATE_NETWORK_CIDRS = [
```
kerin: This should be a variable.

infra/terraform/modules/auth0/rules/add-group-claim-to-token.js
```diff
@@ -0,0 +1,43 @@
+/*
+ * This rule is managed by Terraform, DO NOT EDIT!
+ */
+function (user, context, callback) {
+
+  var targeted_clients = [
+    'p4L2qRcSgWyqHjoHanJ4QyhWL1iX612i'
```
kerin: This should be derived from the Terraform resource that creates this client.

infra/terraform/modules/auth0/rules/add-group-claim-to-token.js
```diff
@@ -0,0 +1,43 @@
...
+  ].indexOf(context.clientID) !== -1;
+
+  var github_identity = _.find(user.identities, {connection: 'github'});
+
+  if (github_identity && targeted_clients) {
+    // For custom claims, you must define a namespace for oidc compliance.
+    // See https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims
+    var namespace = 'https://api.dev.mojanalytics.xyz/claims/';
```
kerin: The domain or env name here should be passed in as a var.

infra/terraform/modules/auth0/rules/whitelist-github-orgs.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * This rule is managed by Terraform, DO NOT EDIT!
+ */
+function (user, context, callback) {
+  var request = require('request');
+
+  var whitelist = ['moj-analytical-services']; // authorized github orgs
```
kerin: Should be a variable.

infra/terraform/modules/auth0/rules/whitelist-google-domains.js
```diff
@@ -0,0 +1,22 @@
+/*
+ * This rule is managed by Terraform, DO NOT EDIT!
+ */
+function (user, context, callback) {
+  var whitelist = ['digital.justice.gov.uk']; //authorized domains
```
kerin: Should be a variable.

infra/terraform/platform-base/main.tf
```diff
@@ -29,12 +29,25 @@ module "cluster_dns" {
...
 }
 
+module "auth0" {
+  source = "../modules/auth0"
+
+  auth0_api_client_id          = "${var.auth0_api_client_id}"
+  auth0_api_client_secret      = "${var.auth0_api_client_secret}"
+  auth0_tenant_domain          = "${var.oidc_provider_domain}"
+  aws_account_id               = "${var.aws_account_id}"
+  env                          = "${terraform.workspace}"
```
kerin: Minor thing, but I always put `env` at the top of the variable list.

infra/terraform/modules/auth0/main.tf
andyhd: Ah - I was stuck on trying to make `auth0_rules_config` a variable and not being able to have interpolated values. When I switched to `locals`, I didn't realise I could use a literal map again :facepalm:

infra/terraform/modules/auth0/main.tf
andyhd: Yeah, this should have been a `locals` entry, but there's no reason it shouldn't be overridable.

infra/terraform/modules/auth0/outputs.tf
andyhd: Yep

infra/terraform/modules/auth0/rules/Multifactor-Google-Authenticator-Do-Not-Rename.js
andyhd: As discussed. This whole label should be passed in, really - if this module is to be usable by people other than us. Easiest way is with a rule config key/val pair, accessed in JS with `configuration[KEY]`

infra/terraform/platform-base/main.tf
andyhd: why?

infra/terraform/modules/auth0/main.tf
andyhd: Is the implication here that `inputs.tf` should be named `variables.tf`?

infra/terraform/platform-base/main.tf
kerin: Just consistency - pretty much every module has that passed in, so I put it at the top so you can easily see that there's a common identifier used by each module.

infra/terraform/modules/auth0/rules/Multifactor-Google-Authenticator-Do-Not-Rename.js
kerin: tbh I wouldn't worry too much about this being reusable in other contexts - the setup here is pretty specific to this product.

infra/terraform/modules/auth0/main.tf
kerin: Hadn't noticed there was an `inputs.tf`, but yes - all other modules (hopefully) use `variables.tf` instead.

infra/terraform/modules/auth0/main.tf
```diff
@@ -0,0 +1,93 @@
...
+  ]
+}
+
+locals {
+  computed_config = {
+    AWS_ACCOUNT_ID          = "${var.aws_account_id}"
+    ENV                     = "${var.env}"
+    KUBECTL_OIDC_CLIENT_ID  = "${auth0_client.kubectl-oidc.id}"
+    OIDC_CLAIMS_NAMESPACE   = "https://api.${var.env}.${var.root_domain}/claims/"
+    MFA_ENABLED_CONNECTIONS = "${auth0_connection.github.name},google-oauth2"
```
andyhd: Not sure if we want to try to set up the Google OAuth2 connection with Terraform too - this needs to change if we do.

kerin: I'd say yes, assuming we can. That could be a later PR though of course.

djsd123:
@andyhd, I was wondering whether we could take this opportunity to rename the app `kubectl-oidc` to something like `k8s-api` or `kubernetes-api`?

andyhd:
Closing for now - managing client connections in Auth0 with Terraform is not straightforward and we may need to fork the Auth0 provider plugin to enable our (admittedly unusual) use case.

davidread:
I'm not making a point, simply just summarizing for myself the slack comments on 16th Jan about this problem:

When there's a new webapp, Concourse sets up the auth for it - it creates a new Auth0 Application (Client) and "enables" GitHub Social Connection for it. But Auth0 Terraform records that "enabling" against the GitHub Social Connection, rather than the Application/Client. The problem being that when we run `terraform apply` it’d disable all the webapps from the GitHub Social Connection, so you couldn't login to any webapps.

PR #245, opened 2019-01-07
Bump pyyaml version to 3.13 to fix security hole in 3.11

andyhd:


infra/terraform/modules/log_pruning/prune_logs/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[install]
+prefix=
```
andyhd: I found I needed this to make `build.sh` run without errors - it may not be necessary

xoen: I'm not familiar with this `setup.cfg` so I'm not sure.
I had a quick look around and [I've found this](https://github.com/pypa/pipenv/issues/2444). According to one of the comments:

> For the record, you can't use pip install with a setup.cfg file that is configured like this.

The [`build.sh` script](https://github.com/ministryofjustice/analytics-platform-ops/blob/master/infra/terraform/modules/log_pruning/prune_logs/build.sh#L3) is running `pip install` so maybe we don't need this `setup.cfg`?

Maybe our other Python ninja @r4vi is familiar with `pipenv`/`setup.cfg`.

andyhd: `build.sh` doesn't use pipenv

andyhd: can you run build.sh without the setup.cfg? if so, i'll remove it

r4vi: is this a lambda function? setup.cfg is telling pip to install the packages in the current directory so I guess it makes sense to have it

r4vi: setup.cfg is a generic python packaging thing, not really related to `pipenv`

r4vi: I don't see how this `build.sh` would work if you didn't have it :confused:

xoen: @r4vi uh, you must be right. I don't remember having to add a `setup.cfg` when I was working at some Python lambda functions months ago 🤷‍♂️ .

I know that lambda deployments are very rudimental and the runtime environment expects a zip file with all the dependencies.
The downloading of the dependencies happens via the `build.sh` script and the terraform's `archive_file` take care of creating the zip file. But as I said, I'm not familiar with `pipenv` so maybe this `setup.cfg` is necessary for the lambda function to run?

xoen: For the record (and our future selves): The `build.sh` is cd-ing in the lambda function directory and then passing the `-t .` flag (`--target`) to pip3 which cause the packages to be installed in the current directory (which is what we want):

```bash
#!/usr/bin/env bash
cd $(dirname "$0")
pip3 install -r requirements.txt --upgrade -t .
```

This means that we shouldn't need the `setup.cfg`.
The error may be caused by and old version of pip3 which doesn't support the `--target` flag.

PR #246, opened 2019-01-07
Upgrade to PyYAML 4.2b1 to fix vulnerability

andyhd:
Upgrade to version which actually fixes vulnerability

### DIFF
--- a/infra/terraform/modules/log_pruning/prune_logs/requirements.txt
+++ b/infra/terraform/modules/log_pruning/prune_logs/requirements.txt
@@ -1,3 +1,3 @@
 certifi==2016.8.8
 elasticsearch-curator==4.0.6
-PyYAML==3.13
+PyYAML>=4.2b1,<5


infra/terraform/modules/log_pruning/prune_logs/requirements.txt
```diff
@@ -1,3 +1,3 @@
 certifi==2016.8.8
 elasticsearch-curator==4.0.6
-PyYAML==3.13
+PyYAML>=4.2b1
```
r4vi: I'd pin it to a known good version rather than `>=4.2b1` latest I can [see](https://pypi.org/project/PyYAML/#history) is `==4.2b4` but it feels weird to pin to a `b` version so maybe do something like `>=4.2b1,<5` so we don't inadvertently get moved to the next version

andyhd: Sure. I just put in what the vulnerability alert said.

PR #248, opened 2019-01-09
Give Atlantis IAM role permissions to create lambda functions and lambda IAM role

kerin:
Part of testing/evaluating Atlantis.


infra/terraform/global/atlantis.tf
```diff
@@ -10,12 +10,28 @@ data "aws_route53_zone" "global" {
...
+    resources = ["*"]
+    effect    = "Allow"
+  }
+}
+
+resource "aws_iam_policy" "atlantis_terraform" {
+  name   = "atlantis-terraform"
+  policy = "${data.aws_iam_policy_document.atlantis_terraform.json}"
+}
+
```
djsd123: Where is this policy being used?

infra/terraform/modules/atlantis_iam_role/main.tf
```diff
@@ -0,0 +1,84 @@
...
+# The Atlantis module outputs the role ARN, but Terraform resources
+# take the role name as an argument, not ARN
+locals {
+  ecs_role_arn_parts = "${split("/", var.atlantis_ecs_role_arn)}"
+  ecs_role_name      = "${element(local.ecs_role_arn_parts, length(local.ecs_role_arn_parts) - 1)}"
+}
+
+data "aws_iam_policy_document" "atlantis_terraform" {
+  statement {
+    actions   = ["ec2:*"]
```
xoen: If I understand correctly for now Atlantis will only be able to perform EC2 actions (e.g. create/destroy "servers" and related stuff) right? (a part from being able do manage TF state S3 bucket)

I guess if we want Atlantis to apply any possible TF changes it will need pretty much root access as you never know what we'll put in Terraform (e.g. DBs, Lambdas, Athena, Glue stuff and god knows what else), is that the case?

PR #250, opened 2019-01-16
Enable Terraform state locking for `global` and create dynamoDB tables for `platform` and `platform-base`

kerin:
Enables [state locking with DynamoDB](https://www.terraform.io/docs/backends/state.html) for `global` resources, and creates DynamoDB tables for state locking in `platform-base` and `platform` resources. State locking will be enabled for `platform` and `platform-base` once the [auth0 provider PR](https://github.com/ministryofjustice/analytics-platform-ops/pull/244) has been merged, as an in-progress `apply` of auth0 provider-managed resources currently means that an `apply` in all branches causes an error due to the missing auth0 plugin. In short, it's better to wait for the auth0 provider changes to be merged before proceeding.

### DIFF
--- a/infra/terraform/global/main.tf
+++ b/infra/terraform/global/main.tf
@@ -1,8 +1,9 @@
 terraform {
   backend "s3" {
-    bucket = "terraform.analytics.justice.gov.uk"
-    key    = "base/terraform.tfstate"
-    region = "eu-west-1"
+    bucket         = "terraform.analytics.justice.gov.uk"
+    key            = "base/terraform.tfstate"
+    region         = "eu-west-1"
+    dynamodb_table = "terraform-global"
   }
 }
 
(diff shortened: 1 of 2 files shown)


infra/terraform/global/main.tf
```diff
@@ -1,8 +1,9 @@
 terraform {
   backend "s3" {
-    bucket = "terraform.analytics.justice.gov.uk"
-    key    = "base/terraform.tfstate"
-    region = "eu-west-1"
+    bucket         = "terraform.analytics.justice.gov.uk"
+    key            = "base/terraform.tfstate"
+    region         = "eu-west-1"
+    dynamodb_table = "terraform-global"
```
davidread: Doesn't this need to be `terraform-platform-base` to match the table name above?

kerin: No - there's a DynamoDB table required for each Terraform state object/resource set - `global`, `platform-base` and `platform`. Global TF uses `terraform-global`, which needs to be created manually as it's the very first thing to run, and the global resources then create the two DynamoDB tables for `platform-base` and `platform`.

davidread: Makes sense. Please do add the creation of the dynamodb table with the other awscli commands in this section: https://github.com/ministryofjustice/analytics-platform-ops#creating-global-aws-resources-and-preparing-terraform-remote-state

PR #254, opened 2019-01-22
Scale highmem from 0

djsd123:
Cluster Autoscaler was missing some of the permissions to scale an autoscaling
group from 0.


djsd123:
Completely forgot about this one 🤔.  Thanks

PR #266, opened 2019-03-19
Script to add `host` label to tools k8s resources

xoen:
This is to simplify unidling logic by leveraging labels.

### DIFF
--- a/scripts/set-tools-host-labels.sh
+++ b/scripts/set-tools-host-labels.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# Add the `host` label to all RStudio/JupyterLab resources
+
+
+set -e
+
+
+if [[ -z "${APP}" ]]; then
+  echo "Please specify APP. It can be 'rstudio' or 'jupyter-lab' at the moment."
+  exit 1
+fi
+
+
+NAMESPACES=`kubectl get ns | grep user- | cut -f1 -d' '`
+for NAMESPACE in $NAMESPACES; do
+  echo "NAMESPACE='$NAMESPACE'"
+
+  # Get host from Ingress resource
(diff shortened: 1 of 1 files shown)


xoen:
**NOTE**: The bit that patch jobs is currently failing for some reason I don't get (yet!).

For example:

```bash
$ kubectl --namespace=user-alice patch job jupyter-lab-alice-jupyt --type='json' -p="[{\"op\": \"add\", \"path\": \"/spec/template/metadata/labels/host\", \"value\": \"alice-jupyter-lab.tools.example.com\"}]"

The Job "jupyter-lab-alice-jupyt" is invalid: spec.template: Invalid value: core.PodTemplateSpec{ [...]
```

Odd. Even trying to add the label manually by editing the job gives the same error 🤷‍♂️

scripts/set-tools-host-labels.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+# Add the `host` label to all RStudio/JupyterLab resources
+
+
+set -e
```
djsd123: You're handling envvars not being set so won't suggest `set -u`.  Maybe `set -o pipefail` might be useful

xoen:
Regarding helm, I didn't think too much about it but I think there are two main reasons why I went for the Bash script instead:
1. I don't want to mass upgrade all tools just to add `host` labels. As you know some users have "special sauce" images and mass upgrade using helm would bring all RStudio/JupyterLab to the latest versions in helm. This is not ideal IMHO.
2. Helm is quite peculiar and in this specific case I prefer to deal directly with `kubectl` instead, this means I don't end up with a error halfway a migration because of some Helm "design decision".

I hope it makes sense.

xoen:
@djsd123 Thanks 🙏 

```
APP="jupyter-lab" ./scripts/set-tools-host-labels.sh
APP="rstudio" ./scripts/set-tools-host-labels.sh
```

PR #271, opened 2019-04-08
Remove OIDC migration IAM role

davidread:
TemporaryForOIDCMigration IAM role was introduced in:
https://github.com/ministryofjustice/analytics-platform-ops/pull/124
and according to that it should no longer be needed.

It gives cpanel a concerning level power in AWS.

### DIFF
--- a/infra/terraform/modules/control_panel_api/role.tf
+++ b/infra/terraform/modules/control_panel_api/role.tf
@@ -147,17 +147,6 @@ resource "aws_iam_policy" "control_panel_api" {
       "Resource": [
         "arn:aws:iam::${var.account_id}:*"
       ]
-    },
-    {
-      "Sid": "TemporaryForOIDCMigration",
-      "Effect": "Allow",
-      "Action": [
-        "iam:GetRole",
-        "iam:UpdateAssumeRolePolicy"
-      ],
-      "Resource": [
-        "arn:aws:iam::${var.account_id}:role/${var.env}_user_*"
-      ]
     }
   ]
 }


xoen:
If you checked and the `iam:GetRole` is not needed and you tested it and it works go for it 👍 

@andyhd is working on the Control Panel so he should look at this as well.

davidread:
@xoen I was hoping you'd indicate that this was indeed no longer needed, since you wrote the original PR ;-) I'll test today.

xoen:
Sorry morning was crazy/taken by management stuff.

I can't really remember why I didn't clean it up after, I vaguely remember having a conversation with someone and agreeing to leave it there for the time being for some reason.

I haven't worked on CP for a while so hard to tell for sure but my main worry was that the `iam:GetRole` was necessary to read the inline policies (to update the permissions).

But it seems like that is also covered by the [`CanReadRolesInlinePolicies` statements](https://github.com/ministryofjustice/analytics-platform-ops/blob/0f86395f196880673a1cd974ba60899000bf69eb/infra/terraform/modules/control_panel_api/role.tf#L119) (as it should) - so that should be fine. Still worth talking with @andy as he's working on CP - I don't imagine this would cause him any trouble but have a quick chat doesn't harm.

andyhd:
This is one part of the control panel that I've left untouched, so I can't speak to the need for this permission. Looking at the code, I don't see this permission being used, so I would say it's fine to remove it. Looks like it was a temporary requirement in any case.

andyhd:
I imagine @xoen left it in because he hates YAGNI 😛

PR #274, opened 2019-06-03
Add script to create Auth0 additional connections

r4vi:
## What

Document how to create additional auth0 connections (without using `terraform`).
This also adds a script to run, which will create the defined connections
instead of documenting how to do this manually in the auth0 console.

related config change: https://github.com/ministryofjustice/analytics-platform-config/pull/143

cc @sh31 because I know you're working on something similar, maybe this can be removed if your way works for creating connections


r4vi:
[Login via NOMIS for apps](https://trello.com/c/9VqGIvrI/226-login-via-nomis-for-apps)

PR #277, opened 2019-06-06
SoftNAS: Add disk

djsd123:
Extending SoftNAS storage pool and importing into terraform state.


infra/terraform/modules/user_nfs_softnas/ebs.tf
```diff
@@ -114,3 +126,11 @@ resource "aws_volume_attachment" "softnas_vol5" {
 
   count = "${var.num_instances}"
 }
+
+resource "aws_volume_attachment" "softnas_vol6" {
```
r4vi: :money_with_wings:

PR #280, opened 2019-07-05
Add parameter permissions to control panel

s-block:
Gives the control panel ability to create parameters

## What

Changed the IAM policy to add statement for creating Params

### DIFF
--- a/infra/terraform/modules/control_panel_api/role.tf
+++ b/infra/terraform/modules/control_panel_api/role.tf
@@ -164,6 +164,16 @@ resource "aws_iam_policy" "control_panel_api" {
       ]
     },
     {
+      "Sid": "CanListRoles",
+      "Effect": "Allow",
+      "Action": [
+        "iam:ListRoles"
+      ],
+      "Resource": [
+        "arn:aws:iam::${var.account_id}:role/*"
+      ]
+    },
+    {
       "Sid": "TemporaryForOIDCMigration",
       "Effect": "Allow",
       "Action": [
--- a/infra/terraform/modules/control_panel_api/role.tf
+++ b/infra/terraform/modules/control_panel_api/role.tf
@@ -149,6 +149,21 @@ resource "aws_iam_policy" "control_panel_api" {
       ]
     },
     {
+      "Sid": "CanCreateAndDeleteSSMParameters",
+      "Effect": "Allow",
+      "Action": [
+        "ssm:PutParameter",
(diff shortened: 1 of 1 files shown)


xoen:
Don't know if we want to keep prefixing resource names by environment to logically separate them, e.g `dev` CP has only permission to touch `"arn:aws:ssm::${var.account_id}:parameter/dev_*"` params and `alpha` CP has only permission to touch `"arn:aws:ssm::${var.account_id}:parameter/alpha_*"` (it would avoid potential problems).

Otherwise looks good from the little I know about AWS parameter store 👍

infra/terraform/modules/control_panel_api/role.tf
```diff
@@ -148,6 +148,21 @@ resource "aws_iam_policy" "control_panel_api" {
...
+      "Action": [
+        "ssm:PutParameter",
+        "ssm:DeleteParameter",
+        "ssm:GetParameterHistory",
+        "ssm:GetParametersByPath",
+        "ssm:DeleteParameter",
+        "ssm:DeleteParameters"
+      ],
+      "Resource": [
+        "arn:aws:ssm::${var.account_id}:parameter/*"
```
r4vi: good point by @xoen about namespacing the params

infra/terraform/modules/control_panel_api/role.tf
```diff
@@ -148,6 +148,31 @@ resource "aws_iam_policy" "control_panel_api" {
...
       ]
     },
+    {
+      "Sid": "CanCreateAndDeleteSSMParameters",
+      "Effect": "Allow",
+      "Action": [
+        "ssm:PutParameter",
+        "ssm:DeleteParameter",
+        "ssm:GetParameterHistory",
+        "ssm:GetParametersByPath",
```
andyhd: Do we need `GetParameterHistory`? And do we not need `GetParameter`?

r4vi:
I would've thought cpanel should only set them and not be able to retrieve
the value

infra/terraform/modules/control_panel_api/role.tf
r4vi: I would've thought cpanel should only set them and not be able to retrieve
the value

s-block: Yeah, there was something about listing in boto which required history. But no, we don't want to get the parameter in control panel

PR #284, opened 2019-07-11
Add get parameter user for concourse

s-block:
Allows concourse to get parameter to add to webapp

## What

Added module to create user that can get and decrypt parameters


infra/terraform/modules/user_get_parameter/iam.tf
```diff
@@ -0,0 +1,49 @@
...
+
+  statement {
+    actions = [
+      "kms:Decrypt",
+    ]
+
+    effect = "Allow"
+
+    resources = [
+      "arn:aws:kms:::key/*",
```
xoen: Do we prefix the keys with the environment as well?

PR #286, opened 2019-07-22
Auth0 connections: Fetch user profile script will now store access token in profile

r4vi:
## What

This stores the user accessToken in the user profile. This is to allow apps to make calls to the nomis api gateway using the auth token received as part of the oauth2 login flow

https://trello.com/c/ZZ4Cs0tI/250-nomis-login-for-sdt


r4vi:
Nomis doesn't provide an `email` field but many parts of our platform rely on there being one. Especially the USER_EMAIL header. I've made a synthetic `email` for nomis users so that this can be passed through to webapps

davidread:
FYI @obutterbach

obutterbach:
Thank you @davidread

PR #292, opened 2019-09-03
Add a module creating Audit logs policy attached to instance role

obutterbach:


obutterbach:
Missing S3 plugin from `gcr.io/google-containers/fluentd-elasticsearch`, this role is not required anymore

PR #293, opened 2019-09-04
Allow Control Panel to set public access block

andyhd:
When creating S3 Buckets in Control Panel, it needs to set public access block by default


obutterbach:
👍

PR #297, opened 2019-11-13
Increase Alpha Node ASG

djsd123:
[Trello](https://trello.com/c/MOz23A2n)

Increasing Alpha node instance group from `6` to `25` for minimum and
maximum from `25` to `50`.

In reality, values were min:`9` and max:`50`.

**Deployed**


djsd123:
Cheers @s-block

PR #304, opened 2020-01-17
ebs_volumes: avoid useless volumes when not production

xoen:
SoftNAS in `alpha` has 2 instances and more volumes than `dev`.
Before this `terraform apply` would try to create vol6 and vol7
in `dev` which is a waste of money.

It's a bit ugly but Terraform is the way it is.

Regarding the `element(concat(..., list("")), n)` it's inspired
by this GitHub issue/comment:

https://github.com/hashicorp/terraform/issues/11210#issue-200774446


xoen:
Thanks @sh31 ☺️ - more PRs coming soon

PR #306, opened 2020-01-20
ebs_snapshots: Add tags to DLM Policy resources

xoen:
This is for consistency and it may help future team members to
determine which environment they're snapshotting, if they're used in
production, etc...

Also added `type` to variables (but interestingly Terraform doesn't have a `boolean` type 🤷‍♂ )

**NOTE**: `aws_dlm_lifecycle_policy`'s support for tags was added in
`aws` provider version `2.37.0+`:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2370-november-18-2019

Part of ticket: https://trello.com/c/4YSbvgFu

### DIFF
--- a/infra/terraform/modules/ebs_snapshots/README.md
+++ b/infra/terraform/modules/ebs_snapshots/README.md
@@ -25,9 +25,11 @@ Parameters
 | ------------------------------------ | ------------ | ----------------------------------------- |
 | `name`                (**Required**) | `string`     | The common name given to resources        |
 | `target_tags`         (**Required**) | `map`        | The DLM will take a snapshot of all EBS volumes with any of these target tags |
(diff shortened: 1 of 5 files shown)


xoen:
FYI @sh31 applied in `dev` and looks good.

xoen:
FYI: I've also added the `env` in the DLM policy description which seems to be the only field shown in the AWS Console: https://github.com/ministryofjustice/analytics-platform-ops/pull/306/files#diff-f18ae658c6fa837613a4a29d6ca4b57dR2

sh31:
Alpha plan looks good to me

xoen:
Thanks @sh31 - I'll apply to `alpha` 💥

PR #308, opened 2020-01-28
Added cloudwatch_alerts module to monitor SoftNAS

xoen:
Send email if CPU spikes over 80% for 2 five minutes periods in the last
15 minutes.

Unfortunately the `aws_sns_topic_subscription` Terraform resource
doesn't support the `email` protocol:

> Unsupported protocols include the following:
> - email -- delivery of message via SMTP
> - email-json -- delivery of JSON-encoded message via SMTP
>
> These are unsupported because the endpoint needs to be authorized and
> does not generate an ARN until the target email address has been
> validated. This breaks the Terraform model and as a result are not
> currently supported.

Apparently this is because of the manual click to subscribe required.

See: https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html

For this reason the module creates a CloudFormation stack which creates
SNS topic. The CloudWatch alarm then uses this SNS topic created by
this stack to send the email notifications.

Ticket: https://trello.com/c/wI7T79OV

### DIFF
--- a/infra/terraform/modules/cloudwatch_alerts/README.md
+++ b/infra/terraform/modules/cloudwatch_alerts/README.md
@@ -1,0 +1,38 @@
+# CloudWatch Alerts
+
+##### Terraform module to enable CloudWatch alerts
+
+Sends alerts when CPU usage goes above ${cpu_threshold} percent.
+
+Usage
+-----
+
+Create the CloudWatch alerts for the specified EC2 instances
+
+```hcl-terraform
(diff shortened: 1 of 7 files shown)


xoen:
FYI: This article explains the weird CloudFormation workaround to the lack of Terraform support for `email` protocol for `aws_sns_topic_subscription`: http://aws-cloud.guru/terraform-sns-topic-email-list

PR #311, opened 2020-02-21
S3 bucket where archived buckets data goes

xoen:
## What

Creates the `${ENV}-archived-buckets-data` S3 bucket.
Data from other buckets will be moved here when they're archived.

Key features of this S3 bucket:
- is private
- bucket has versioning enabled
- is encrypted
- has lifecycle rule which deletes objects after 6 months
  (configurabe)
- has public access blocks in place

The module also creates the IAM role that will be assumed by the buckets archiver.

## How to review

1. check `terraform plan`
2. ???
3. Profit!

## Other
I've also used this opportunity to move out of the modules into a `tags` variable the standard tags we use.

We want to add standard tags to as many resources as possible.
Ideally all AWS resources should have standard tags to be able to easily
determine who's the owner of a resouce, which environment is using it,
if it's used in production, component using the resource (e.g. "SoftNAS",
"Control Panel", etc...).

Most of the resources will have the same tags or the same tags plus
few additional ones.

For this reason I've moved these standard tags outside the modules into
a variable in the `platform` Terraform stack.

Some modules were refactored to:
1. require mandatory `tags` variable
2. use tags as they passed in with few exceptions where they need to
add a `Name` tag or something similar
3. READMEs and example was updated accordingly
4. some variables were removed as were only used in the tags

The client code which instanciates the module would now pass
in `${var.tags}` and sometimes adding a tag when necessary (e.g.
the `softnas_monitoring` module adds the `component=SoftNAS` tag to the
standard tags)

Above changes are refactoring which wouldn't cause `terraform apply` to
change anything.

In addition to those `user_nfs_softnas` module is now tagging the
EC2 resources (EC2 instances and network interfaces) with the same
standard tags. These **will** cause `terraform apply` to change those
resources as these tags will need to be added but this shouldn't in
theory cause any problems as this is a metadata change.

## Ticket
Part of ticket: https://trello.com/c/6VdNPbzz


xoen:
@sh31 FYI: I tweaked a couple of minor things:
- https://github.com/ministryofjustice/analytics-platform-ops/pull/311/commits/885c9f953e5cdd9bfc006858234a6d92819ea4c7
- https://github.com/ministryofjustice/analytics-platform-ops/pull/311/commits/1b59dbb1f49d519b3759171293c45326b8812fc3

and applied to `dev` (only the `buckets_archiver` module for now).

I'm still building the `buckets_archiver` so I imagine I may need to tweak the permissions further, so not merging this yet.

xoen:
@dhirajnarwanimoj / @sh31 I've tweaked the CP IAM role so that CP can read existing tags for a bucket: https://github.com/ministryofjustice/analytics-platform-ops/pull/311/commits/21d6338c1c19f6b6c72a351a052c2fef81c77ec7

Let me know if it's good to `terraform apply` - maybe you could double check the output of `terraform plan` to see if you see anything out of the ordinary/wrong 🙏

PR #314, opened 2020-03-25
Create a shell script that updates jupyter, rstudio and other user helm charts

dhirajnarwanimoj:
## What

Added variable $HELM_CHART to specify any helm chart you like to update for a user
Added upgrade --install to upgrade HELM Charts rather than remove and then install

## How to review

1. Change script locally to point to your helm chart repo for jupyter, rstudio and a helm chart you want to update (e.g config-user)
2. run script on dev and see it update your helm charts with the latest Chart version
3. rollback to previous release to make sure rollback works

https://trello.com/c/P7BRQJWC/333-strip-out-ipython-notebook-output-using-a-git-hook-for-all-users


scripts/helm_redeploy_all_users.sh
```diff
@@ -0,0 +1,96 @@
...
+		*	) echo "Only dev or alpha are acceptable answers"; exit;;
+	esac
+done
+
+CHART="../../analytics-platform-helm-charts/charts/${HELM_CHART}"
+RSTUDIO_CHART="../../analytics-platform-helm-charts/charts/rstudio"
+JUPYTER_CHART="../../analytics-platform-helm-charts/charts/jupyter-lab"
+RSTUDIO_CONFIG_FILE="../../analytics-platform-config/chart-env-config/${ENVIRONMENT}/rstudio.yml"
+JUPYTER_CONFIG_FILE="../../analytics-platform-config/chart-env-config/${ENVIRONMENT}/jupyter.yml"
+# USERNAME=$(helm list --max 10000 | grep "${HELM_CHART}" | awk '{print $1}' | sed "s/${HELM_CHART}-//g")
```
davidread: The script is called "redeploy_all_users" so I think you should uncomment the next line instead of this one, for the version you commit.

scripts/helm_redeploy_all_users.sh
```diff
@@ -0,0 +1,96 @@
...
+
+	set -x
+
+	while IFS= read -r u; do
+		
+		kubectl delete deployment -l app=jupyter-lab -n user-$u
+
+		kubectl delete deployment -l app=rstudio -n user-$u
+
+		if [[ $? -ne 0 ]]; then
```
davidread: Doesn't this if only react to the rstudio deletion, and a failed jupyter one is ignored?

scripts/helm_redeploy_all_users.sh
```diff
@@ -0,0 +1,96 @@
...
+
+			kubectl delete deployment -l app=jupyter-lab -n user-$u
```
davidread: It would be good to understand the circumstance that lead to needing a retry - I suggest a code comment

scripts/helm_redeploy_all_users.sh
```diff
@@ -0,0 +1,96 @@
...
+			kubectl delete deployment -l app=rstudio -n user-$u
+		fi
+
+		RELEASE_NAME=$HELM_CHART-$u
+
+		JUPYTER_RELEASE_NAME=jupyter-lab-$u
+
+		RSTUDIO_RELEASE_NAME=rstudio-$u
+
+		# add -f $CONFIG_FILE if you want to specify a config file
```
davidread: Maybe you don't need this comment, since you're already specifying a `-f` value?

scripts/helm_redeploy_all_users.sh
```diff
@@ -0,0 +1,96 @@
...
+
+username_file
+
+redeploy() {
+
+	set -x
+
+	while IFS= read -r u; do
+		
+		kubectl delete deployment -l app=jupyter-lab -n user-$u
```
davidread: Maybe explain why you need to delete these deployments before upgrading their charts. I'd assume this would not be necessary, so it would be good to understand the reasoning.

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: I'll change the name then to say per user

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: Yes because when i update the helm chart it will create new pods and not delete the old ones

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: Yes if everything doesn't complete then it will retry

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: Timeshot / TLS handshake issues

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: Ok

scripts/helm_redeploy_all_users.sh
davidread: Really, upgrading a helm chart doesn't delete its previous deployments?! Either it's just me that didn't know this, or a snarky code comment along this line is needed ;-)

xoen: @dhirajnarwanimoj the `--recreate-pods` flag may help with this: https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback

scripts/helm_redeploy_all_users.sh
dhirajnarwanimoj: We'll leave this all users and we need to roll this out to list of users once we finish testing with the beta users

davidread:
I think the git-crypt stuff is separate to the helm chart update script. Would you mind moving that to a separate PR? And I think the git-crypt stuff deserves its own .txt file.

dhirajnarwanimoj:
> I think the git-crypt stuff is separate to the helm chart update script. Would you mind moving that to a separate PR? And I think the git-crypt stuff deserves its own .txt file.

What do you mean separate .txt file. Do you mean README?

damacus:
You might want to rebase this PR or start again :)

PR #316, opened 2020-04-15
auth0connections - add hmpps-auth connection and refactor

davidread:
This PR changes the way to define connections. Previously you defined a template and it referred to a piece of config.
But we found that we had lots of similar connections with the same template but different config (HMPPS Auth for app1 (dev), HMPPS Auth for app2 (prod), etc)
With this PR, connections are now defined in the config and refer to a template (i.e. in connections directory).
This allows us to reuse the same connection template for lots of different connections.
And this also lets us have a different set of connections on dev, compared to alpha. (To make dev connections available on both, but not prod connections.)

This change goes with an update to: analytics-platform-config/chart-env-config/dev/auth0connections.yaml
See the PR: https://github.com/ministryofjustice/analytics-platform-config/pull/203
Each connection has the new connection value:

    connection_template: hmpps-auth

And some small changes:
* added a 'delete connection' command
* 'remote' command has a new parameter just to print the names of connections


xoen:
I'm pretty sure I've never seen this code before. I'll have a look and see if I can grasp some of it 😁  (didn't see the ticket link but from description this should be solely a refactor)

scripts/auth0connections/auth0connections.py
```diff
@@ -29,26 +30,44 @@ def get_client():
...
+         means use the template in: ./connection_templates/hmmps-auth/
+    '''
     connections = {}
-    connections_root = Path(__file__).cwd() / Path("connections")
-    connection_dirs = [entry for entry in connections_root.iterdir() if entry.is_dir()]
-
-    for connection in connection_dirs:
-        connection_name = connection.stem
-        scripts = connection.glob("*.js")
+    template_root = Path(__file__).cwd() / Path("connection_templates")
```
xoen, line 41: I see the templates directory was renamed (seem to make sense if these are templates).
The README may need some tweaks as a result of the renaming.

scripts/auth0connections/auth0connections.py
```diff
@@ -65,13 +84,19 @@ def cli(ctx, config_file):
...
     """
     click.echo("Remote connections:")
     client = get_client()
-    click.echo(yaml.safe_dump(client.connections.all()))
+    if names:
+        click.echo(yaml.safe_dump(
+            [c['name'] for c in client.connections.all()]
+        ))
+    else:
+        click.echo(yaml.safe_dump(client.connections.all()))
```
xoen, line 99: Not a big deal but there is some duplication between the `if` and the `else` branches, maybe you could write something like this (just an idea):

```python
[...]
connections = client.connections.all()
if names:
    connections = [ c["name"] for c in connections ]

click.echo(yaml.safe_dump(connections))
```

scripts/auth0connections/auth0connections.py
```diff
@@ -95,13 +120,31 @@ def create(ctx):
...
+@click.argument('name')
+def delete(ctx, name):
+    click.echo(f"Deleting connection {name}")
+    client = get_client()
+    remote_connections = dict((c["name"], c) for c in client.connections.all())
+    try:
+        connection_id = remote_connections[name]['id']
+    except KeyError:
+        click.echo(f"Error: Connection {name} does not exist (remotely)", err=True)
+        sys.exit(1)
```
xoen, line 143: The `try`/`except` is probably overkill here, the name not being there wouldn't be that exceptional.
Probably something like:

```python
if name not in remote_connections:
    click.echo(f"Error: ...")
    sys.exit()
```

seems less scary ☺️ 

Also, I've seen this `dict((key, value) for ...` earlier and I was puzzled, maybe it's just my bad python but it took me a bit to realise it was really building a dictionary from the loop. Any reason to not use something like:

```python
remote_connections = { c["name"]: c for c in client.connections.all() }
```

I don't mind really, it's just that the sight of curly braces is a strong cue for the fact we're building a dictionary which may help slow people like me ☺️ 

Lastly, as far as I can tell we're building a dictionary her to get the `id` of the connection to delete right? If that's the case maybe you could avoid building this dictionary and just search for the client with the given name. 
But to be fair I'm not sure that may simplify the code much as you would still have to have a for with an `if` and a `break` - so maybe build a dictionary is less ugly. 
Unless we have a huge amount of connections it wouldn't make much difference memory-wise so probably building the dictionary is good enough.

scripts/auth0connections/auth0connections.py
```diff
@@ -29,26 +30,44 @@ def get_client():
...
+
+    for connection_name in config:
+        connection = config[connection_name]
+        connection['name'] = connection_name
+        template_name = connection['connection_template']
+        try:
+            template_path = template_dirs[template_name]
+        except KeyError:
+            print(f'template_name: "{template_name}" is specified in the config, but no such template exists in {template_root}')
+            sys.exit(1)
```
xoen, line 52: again potentially you may change to not use `try`/`except` and simply check if the key is in the dictionary which looks less drastic. Something like:

```python
if template_name not in template_dirs:
    print([...])
    sys.exit(1)

```

scripts/auth0connections/auth0connections.py
```diff
@@ -29,26 +30,44 @@ def get_client():
...
-    connection_dirs = [entry for entry in connections_root.iterdir() if entry.is_dir()]
-
-    for connection in connection_dirs:
-        connection_name = connection.stem
-        scripts = connection.glob("*.js")
+    template_root = Path(__file__).cwd() / Path("connection_templates")
+    template_dirs = dict((entry.stem, entry) for entry in template_root.iterdir() if entry.is_dir())
+
+    for connection_name in config:
+        connection = config[connection_name]
```
xoen, line 45: from the little I'm understanding this was done in the following line before:

```python
scripts_rendered[name] = template.render(**config.get(connection_name))
```

but not the connection is extracted from the `config` variable here.

I guess you changed the `.get()` with `[]` to make it fail as soon as possible right?

scripts/auth0connections/auth0connections.py
davidread: Good point. I've now updated the readme with this.

scripts/auth0connections/auth0connections.py
davidread: So it would be:
```python
    try:
        connection_id = remote_connections[name]['id']
    except KeyError:
        click.echo(f"Error: Connection {name} does not exist (remotely)", err=True)
        sys.exit(1)
```
vs
```python
if name not in remote_connections:
    click.echo(f"Error: Connection {name} does not exist (remotely)", err=True)
    sys.exit(1)
connection_id = remote_connections[name]['id']
```
I think either is fine. I'm not sure there is any python style guidance on only using try/except when the unhappy path is really 'exceptional', is there? I like try/except because it allows the reader to ignore the unhappy path when skimming.

+1 for using click.echo

I got the impression people preferred `dict((key, value) for ...` over `{key: value for ...}` but will be interested if you can find a reference for this.

I guess these are style things. Maybe @ntoll can suggest from his experience?

scripts/auth0connections/auth0connections.py
davidread:
```
for connection_name in config:
        connection = config[connection_name]
```
We can be sure the key exists because of the previous line, but yes if that changed in future then the `[]` shows intention & failing early 👍

PR #320, opened 2020-05-27
Add a one-shot script to bulk update alpha S3 buckets to be versioned.

ntoll:
## What

This PR complements [this PR in the control panel](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/816) for ensuring S3 buckets are versioned by default and have a life cycle so that non-current versions of files are set to glacier storage after 30 days.

**This script should only be run once, when the control panel PR is merged and deployed.**

The related ticket in Trello is here: https://trello.com/c/g5NFH4Mi/588-s3-bucket-backups-versioning

## How to review

This script was run against a single "alpha" S3 bucket and the end result was verified by @andylightfoot.

### DIFF
--- a/infra/terraform/modules/control_panel_api/role.tf
+++ b/infra/terraform/modules/control_panel_api/role.tf
@@ -40,7 +40,9 @@ resource "aws_iam_policy" "control_panel_api" {
         "s3:CreateBucket",
         "s3:PutBucketLogging",
         "s3:PutBucketPublicAccessBlock",
-        "s3:PutEncryptionConfiguration"
+        "s3:PutEncryptionConfiguration",
+        "s3:PutBucketVersioning",
+        "s3:PutLifecycleConfiguration"
       ],
       "Resource": [
         "arn:aws:s3:::${var.env}-*"
--- a/scripts/version_buckets.py
+++ b/scripts/version_buckets.py
@@ -69,16 +69,40 @@ def is_control_panel_bucket(name, bucket):
             tags = bucket.Tagging()
             tags.load()
             for obj in tags.tag_set:
-                if obj["Key"] == "buckettype" and obj["Value"] == "datawarehouse":
+                if (
+                    obj["Key"] == "buckettype"
+                    and obj["Value"] == "datawarehouse"
+                ):
                     is_valid = True
                     break
         except ClientError as ex:
-            if ex.response.get('Error', {}).get('Code') != 'NoSuchTagSet':
+            if ex.response.get("Error", {}).get("Code") != "NoSuchTagSet":
                 raise
             # It wasn't possible to get the bucket's tags.
             is_valid = False
     logger.info(f"{is_valid} - Bucket {name} is associated with CP")
     return is_valid
+
+
+def get_bucket(name):
+    """
+    Return details of the named bucket (or none, if it isn't a bucket that's
+    valid for the purposes of this script).
+    """
+    bucket_obj = boto3.resource("s3").Bucket(name)
+    if is_control_panel_bucket(name, bucket_obj):
+        versioning = bucket_obj.Versioning()
+        try:
+            lifecycle_conf = bucket_obj.LifecycleConfiguration()
+            lifecycle = lifecycle_conf.rules
+        except ClientError as ex:
+            lifecycle = ""
+        return {
+            "bucket": bucket_obj,
+            "versioning": versioning.status,
+            "lifecycle": lifecycle,
+        }
+    return None
 
 
 def get_buckets(number=0):
@@ -101,24 +125,16 @@ def get_buckets(number=0):
     with click.progressbar(buckets, len(buckets)) as bucket_list:
         for bucket in bucket_list:
             name = bucket["Name"]
-            bucket_obj = boto3.resource("s3").Bucket(name)
-            if is_control_panel_bucket(name, bucket_obj):
-                versioning = bucket_obj.Versioning()
-                try:
-                    lifecycle_conf = bucket_obj.LifecycleConfiguration()
-                    lifecycle = lifecycle_conf.rules
-                except ClientError as ex:
-                    lifecycle = ""
-                result[name] = {
-                    "bucket": bucket_obj,
-                    "versioning": versioning.status,
-                    "lifecycle": lifecycle,
-                }
+            bucket_metadata = get_bucket(name)
+            if bucket_metadata:
+                result[name] = bucket_metadata
     return result
 
 
 @click.group()
-@click.option("--verbose", is_flag=True, help="Comprehensive logging set to stdout.")
+@click.option(
+    "--verbose", is_flag=True, help="Comprehensive logging set to stdout."
+)
 def main(verbose=False):
     """
     A tool to manage the versioning and life cycle of S3 buckets.
@@ -168,20 +184,30 @@ def list(number=0):
 
 
 @main.command()
-@click.option("-n", "--number", required=False, type=int)
+@click.option("-n", "--name", required=False, type=str)
+@click.option("-a", "--amount", required=False, type=int)
 @click.option("-x", "--execute", is_flag=True)
-def update(number=0, execute=False):
+def update(name="", amount=0, execute=False):
     """
-    Update n (or all, if not given) bucket details. Use --execute to make the
-    changes happen, or else a preview will take place.
+    Update either the named bucket or amount number of buckets, or all, if not
+    given all buckets. Use --execute to make the changes happen, or else a
+    preview will take place.
     """
-    buckets = get_buckets(number)
-    for k, v in buckets.items():
-        name = k
-        bucket = v["bucket"]
-        versioning = v["versioning"] == "Enabled"
-        lifecycle = v["lifecycle"]
-        update_bucket(name, bucket, versioning, lifecycle, execute)
+    if name:
+        bucket_metadata = get_bucket(name)
+        if bucket_metadata:
+            bucket = bucket_metadata["bucket"]
+            versioning = bucket_metadata["versioning"] == "Enabled"
+            lifecycle = bucket_metadata["lifecycle"]
+            update_bucket(name, bucket, versioning, lifecycle, execute)
+    else:
+        buckets = get_buckets(amount)
+        for k, v in buckets.items():
+            name = k
+            bucket = v["bucket"]
+            versioning = v["versioning"] == "Enabled"
+            lifecycle = v["lifecycle"]
+            update_bucket(name, bucket, versioning, lifecycle, execute)
 
 
 def update_bucket(name, bucket, versioning, lifecycle, execute):
@@ -199,44 +225,47 @@ def update_bucket(name, bucket, versioning, lifecycle, execute):
         if not VERBOSE:
             click.echo(msg)
         if execute:
-            # TODO: Uncomment to actually make the changes.
-            # v = bucket.Versioning()
-            # v.enable()
+            v = bucket.Versioning()
+            v.enable()
             click.secho("OK", fg="green")
         else:
             click.secho("OK", fg="yellow")
     # Set life cycle rule to send non-current versions of files to glacier
-    # storage after 30 days.
-    lifecycle_id = f"{name}_lifecycle_configuration"
-    msg = f"Setting lifecycle {lifecycle_id} for bucket {name}."
-    logger.info(msg)
-    if not VERBOSE:
-        click.echo("\n\n" + msg)
-    life_cycle = {
-        "Rules": [
-            {
-                "ID": lifecycle_id,
-                "Status": "Enabled",
-                "Prefix": "",
-                "NoncurrentVersionTransitions": [
-                    {"NoncurrentDays": 30, "StorageClass": "GLACIER",},
-                ],
-            },
-        ]
-    }
-    msg = json.dumps(life_cycle)
-    logger.info(msg)
-    if not VERBOSE:
-        click.echo(msg)
-    if execute:
-        # TODO: Uncomment to actually make the changes.
-        # lifecycle_conf = boto3.client("s3").put_bucket_lifecycle_configuration(
-        #    Bucket=bucket_name,
-        #    LifecycleConfiguration=life_cycle
-        # )
-        click.secho("OK", fg="green")
+    # storage after 30 days. Only do this is there is not already a life cycle,
+    # otherwise warn the user.
+    if lifecycle:
+        click.secho(f"Lifecycle already exists for {name}.", fg="red")
     else:
-        click.secho("OK", fg="yellow")
+        lifecycle_id = "lifecycle_configuration"
+        msg = f"Setting lifecycle {lifecycle_id} for bucket {name}."
+        logger.info(msg)
+        if not VERBOSE:
+            click.echo("\n\n" + msg)
+        life_cycle = {
+            "Rules": [
+                {
+                    "ID": lifecycle_id,
+                    "Status": "Enabled",
+                    "Prefix": "",
+                    "NoncurrentVersionTransitions": [
+                        {"NoncurrentDays": 30, "StorageClass": "GLACIER",},
+                    ],
+                },
+            ]
+        }
+        msg = json.dumps(life_cycle)
+        logger.info(msg)
+        if not VERBOSE:
+            click.echo(msg)
+        if execute:
+            lifecycle_conf = boto3.client(
+                "s3"
+            ).put_bucket_lifecycle_configuration(
+                Bucket=name, LifecycleConfiguration=life_cycle
+            )
+            click.secho("OK", fg="green")
+        else:
+            click.secho("OK", fg="yellow")
 
 
 if __name__ == "__main__":
--- a/scripts/version_buckets.py
+++ b/scripts/version_buckets.py
@@ -1,4 +1,14 @@
+#!/usr/bin/env python
 """
+Requires boto3 and click packages to be installed.
+
+1) Create a new virtualenv (python3 -m venv my_venv).
+2) Enable the virtualenv (my_venv/bin/activate).
+3) Install the dependencies: (pip install boto3 click).
+4) Run the command (--help will give you more information).
+
+What does this script do?
+
 Reads all the current buckets, ensures that versioning is switched on for those
 that don't yet have it and sets the life cycle configuration to send
 non-current versions of files to glacier storage after 30 days.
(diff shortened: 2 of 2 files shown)


scripts/version_buckets.py
```diff
@@ -0,0 +1,53 @@
...
+control panel, as a default.)
+"""
+import boto3
+
+
+s3 = boto3.client('s3')
+buckets = s3.list_buckets()
+for bucket_dict in buckets["Buckets"]:
+    bucket_name = bucket_dict["Name"]
+    if not bucket_name.startswith("alpha"):
```
xoen, line 21: 🤔 this only runs on buckets with `alpha-` as prefix but it's still quite wide.
Don't know how else to reduce this tho so I guess will have to do but see next comment.

scripts/version_buckets.py
```diff
@@ -0,0 +1,53 @@
...
+                    "NoncurrentVersionTransitions": [
+                        {
+                            'NoncurrentDays': 30,
+                            'StorageClass': 'GLACIER',
+                        },
+                    ]
+                },
+            ]
+        }
+    )
```
xoen, line 53: Again this script is effecting a large number of S3 buckets, my main concern with this is that it always seem to set (put) the lifecycle regardless of whether the versioning was enabled or not.

For example there are S3 buckets created with terraform (or similar) and they may have a different lifecycle policy and this would change them conflicting with whatever terraform (or similar) wants.

Maybe a safer thing to do would be to only set this lifecycle policy if the versioning is not enabled, at least you reduce the chance to change this on buckets not created by the CP before your addition?

davidread: Yeah, it makes me nervous running such a big job like.How about the script also checks the existing versioning setting and any lifecycle policies. If we run it '--dry-run' we can review the report before doing any changes. And maybe it's overkill, but maybe we need to do a gradual roll out, say 10%, give it a couple of days, then 25%, 50%, 100%, just in case?

scripts/version_buckets.py
ntoll: OK... so what's the way to reliably identify the buckets created by the CP for user data..?

ntoll:
I've added the IAM actions as per Aldo's instructions so the CP has the appropriate permissions. Please confirm these are the right ones / look right to you.

scripts/version_buckets.py
ntoll: Ack.

I think a `--dry-run` flag is good so we can be sure we're identifying the right buckets and transforming them in the correct way. However, once we're happy with the planned changes, I don't see why we can't just do them all at once...? I'd welcome thoughts / advice on this (am I missing something here..?)

infra/terraform/modules/control_panel_api/role.tf
```diff
@@ -40,7 +40,9 @@ resource "aws_iam_policy" "control_panel_api" {
         "s3:CreateBucket",
         "s3:PutBucketLogging",
         "s3:PutBucketPublicAccessBlock",
-        "s3:PutEncryptionConfiguration"
+        "s3:PutEncryptionConfiguration",
+        "s3:PutBucketVersioning",
+        "s3:PutLifecycleConfiguration"
```
xoen, line 45: This should give the CP enough permissions but you never know 😬

xoen: oh, add a trailing comma if you wish (it should be allowed in Terraform right?), it's not really a big deal but it helps when adding new things and with the commit history, etc...

scripts/version_buckets.py
xoen: I'd also move the `put_bucket_lifecycle_configuration()` inside the `if not versioning.status == "Enabled":` to reduce the impact of this, it's kinda of a migration and no point in overwriting this on bucket which already have versioning enabled as they may be have been created by terraform, not Control Panel.

davidread: > I don't see why we can't just do them all at once...?

Because system health is really important to us. We have to make a judgement on what defensive deployment techniques are appropriate or overkill.

To compare, when we deploy software with helm, it does blue-green deploys and we can rollback reliably with one command. When we deploy infrastructure with terraform, the state before and after are stored in git, and again it has the tools to easily and reliably roll-back. But this PR's migration by script is rather more blunt. So whilst there's nothing specific about this change that looks risky, I think some caution is necessary. The risk would be lower if the team had done recent changes of this type.

infra/terraform/modules/control_panel_api/role.tf
ntoll: Normally I would... but I'm following the trailing comma convention I found already in that file... ;-)

scripts/version_buckets.py
xoen: sorry just saw this comment, good question, I thought there wasn't really a way but during standup I thought of something that may work.

so:
- [datawarehouse buckets should have a `buckettype=datawarehouse` label](https://github.com/ministryofjustice/analytics-platform-control-panel/blob/ff8589aa5aa0b7fcc1d96767290f9de74dc1c869/controlpanel/api/aws.py#L209-L210) (did some digging and [this was first introduced ~2 years ago here](https://github.com/ministryofjustice/analytics-platform-control-panel/commit/95adf0d227cedbfacea557d6ed99c4e89c2455b4) AFAICT)
- webapps buckets don't have any labels at the moment (this is also information potentially) but their name starts with `alpha-app-`

So _maybe_ if you combine these two things you _could_ infer that if an S3 Bucket starts with `alpha-app-` or has the `buckettype=datawarehouse` chances are that it was created by the Control Panel.

Now, of course it's just an idea, I can't imagine S3 buckets created beyond the CP to have similar characteristics but of course it's hard to tell.

ntoll: Thanks for the heads up. I'll have a play. :+1:

davidread: Alternatively the control panel displays these buckets, so you could use whatever method it uses to identify them. I believe they will be listed in the db. But the tag might be easiest, if it is indeed reliable.

xoen: Oh yeah @davidread - I guess an alternative is to extract this list from the CP and use it as a whitelist of S3 buckets to migrate, I can't really tell from the top of my head whether this would be simpler than getting this information from AWS (potentially could be simpler) but it's definitely more accurate than that heuristics.

I suppose a benefit of a whitelist is that potentially you could run on few buckets first before pulling the trigger on the whole list.

ntoll: OK, latest push implements this recipe. Please check.

scripts/version_buckets.py
ntoll: Ack.

So what approach to running this script would be most appropriate. For instance I can:

* Grab the names of all the buckets to be changed (from the PoV of the script).
* We confirm these look right (eyeball mk.1)
* We group these buckets and run the script against each bucket at a time.
* After migrating each bucket, and before migrating the next bucket, we confirm (eyeball mk.1) that the migration is as we expect.
* I suggest early groups of buckets are small and contain buckets that belong to us.

Does this help..?

scripts/version_buckets.py
```diff
@@ -18,9 +18,24 @@
...
+    # https://github.com/ministryofjustice/analytics-platform-ops/pull/320#discussion_r433765261
+    update_flag = bucket_name.startswith("alpha-app-")
+    try:
+        tags = bucket.Tagging()
+        tags.load()
+        for obj in tags.tag_set:
+            if obj["Key"] == "buckettype" and obj["Value"] == "datawarehouse":
+                update_flag = True
+                break
+    except:
```
davidread, line 35: A bare except for any particular reason? Are you just wanting to catch KeyError or are there others?

scripts/version_buckets.py
davidread: 👍 Sounds good. 

How about running the script `--dry-run` and pasting the output in this PR, so we can review it's intended actions? It can show for each bucket if it already has versioning and lifecycle config, and it wouldn't need to do any changes.

In terms of doing only some buckets, you could perhaps add a `-n` to do only a limited number of them. My feeling is we can simply sort them in reverse alphabetical order, to leave the covid and nomis ones out of the first batches. Try `-n 1` then `-n 10`, and if there are no shouts of pain after couple of days, do `-n 50`, or something.

scripts/version_buckets.py
ntoll: Good question. This was a matter of convenience. The bucket tagging is complicated, inconsistent and so there were several exceptions that could have been thrown. At this point in the code the only thing that is interesting is to set the flag if, and only if, the conditions on line 32 are met. For all other situations (including throwing an exception) the bucket can be safely ignored.

ntoll:
OK... here's the output from the script that previews the state of all matching s3 buckets. Please check and let me know if there are false positives, or missing buckets..!

[s3states.txt](https://github.com/ministryofjustice/analytics-platform-ops/files/4753366/s3states.txt)

scripts/version_buckets.py
```diff
@@ -0,0 +1,236 @@
...
+LOGFILE = os.path.join(str(Path.home()), "bucket_update.log")
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+logfile_handler = logging.FileHandler(LOGFILE)
+log_formatter = logging.Formatter("%(levelname)s: %(message)s")
+logfile_handler.setFormatter(log_formatter)
+logger.addHandler(logfile_handler)
+
+
+def check_bucket(name, bucket):
```
davidread, line 51: how about calling this method `is_control_panel_bucket`, to make it clearer what it is checking

scripts/version_buckets.py
davidread: A couple of points from PEP8

> A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:).

How about `except Exception:` as a minimum, otherwise at this time a Ctrl-C will cause the script to ignore a bucket but carry on running.

> Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs

What exceptions will `tags = bucket.Tagging()` raise if there is a random network error in the AWS call? In this case we would want to know about it, rather than ignore the bucket.

scripts/version_buckets.py
ntoll: Done in latest push.

ntoll:
I've made sure any exceptions are logged.

scripts/version_buckets.py
```diff
@@ -0,0 +1,236 @@
...
+            for obj in tags.tag_set:
+                if obj["Key"] == "buckettype" and obj["Value"] == "datawarehouse":
+                    is_valid = True
+                    break
+        except:
+            is_valid = False
+    return is_valid
+
+
+def get_buckets(n=0):
```
davidread, line 78: Maybe `number` is clearer than `n`

scripts/version_buckets.py
davidread: I can attest that it is annoying not to be able to use Ctrl-C during `python version_buckets.py list`

scripts/version_buckets.py
```diff
@@ -0,0 +1,236 @@
...
+                try:
+                    lifecycle_conf = bucket_obj.LifecycleConfiguration()
+                    lifecycle = lifecycle_conf.rules
+                except ClientError as ex:
+                    lifecycle = ""
+                result[name] = {
+                    "bucket": bucket_obj,
+                    "versioning": versioning.status,
+                    "lifecycle": lifecycle,
+                }
```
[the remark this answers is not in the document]
ntoll: No, because we only need to check tagging if the bucket doesn't meet the starts-with `alpha-app` requirement (i.e. saving on yet another network request).

ntoll:
@davidread latest commit addresses your comments and also adds logging so we have a record of the decision made for each bucket in `is_control_panel_bucket`.

ntoll:
I want to try to improve our detection of the S3 buckets. Currently the heuristic I use for the name of the bucket is that it starts with `alpha-app`. However, @davidread points out that `alpha-hmpps-ct-tool` is a bucket that the heuristic has missed. Can someone confirm that it should be just a check for starting with `alpha-`..?

xoen:
@ntoll we can't just run this on everything starting with `alpha-` because that would change S3 buckets which are managed by other means (e.g. Terraform), assuming that's was the question.

Regarding the `alpha-hmpps-ct-tool`, I've checked the bucket tags and it has the `buckettype=datawarehouse` tag so not sure why the script missed it.

ntoll:
OK... I've looked into this and the `alpha-hmpps-ct-tool` is correctly listed. I think we're in a state where we know the buckets, we now have to decide how/when to update them.

davidread:
I think the problem with `alpha-hmpps-ct-tool` is not that it is tagged incorrectly, but that I got a connection error. I see from the logs that I get lots of `urllib3.exceptions.NewConnectionError` e.g.:
```ERROR: Could not connect to the endpoint URL: "https://alpha-gluejobutils.s3.eu-west-1.amazonaws.com/?tagging"```

I wonder if we are hitting the API too hard? Or maybe I just have a ropey internet connection.

Whichever, I really think the tags.load() exception handler needs to differentiate between a NoSuchTagSet error which is expected for buckets with no tags, and unexpected errors like connection errors, which should flag the issue to the user.

**Comment by ntoll** [comment]
@davidread aha... I think it might be your flakey internet since I don't get such connection errors in my log. I agree that it's better (more explicit) to trap `ClientError` exceptions. I'll push a fix in a moment.

**Comment by ntoll** [comment]
Deploy Plan

TL;DR a dry run of the following on `dev` followed by a staged rollout on `alpha`.

* Run the script with the `list` option and ensure the expected S3 buckets are listed.
* Run the script for a handful of S3 buckets - monitor these changes.
* Run the script again for a larger sub-section of the buckets - monitor for problems.
* Run the script for all remaining buckets.

Questions: a small number of the buckets already have life cycle specifications associated with them. How should we handle these?

**Comment by davidread** [comment]
@ntoll How about adding numbers like these:
```
* Run the script with the `list` option and ensure the expected S3 buckets are listed.
* Run the script for 20 buckets - monitor these changes for 2 working days
* Run the script for again 40% of the buckets - monitor for problems for 2 working days
* Run the script for all remaining buckets.
```
> Questions: a small number of the buckets already have life cycle specifications associated with them. How should we handle these?

I guess it depends on what they are. You could infer their motivation or ask the bucket owner. On that basis you could decide.

**Comment by xoen** [comment]
FYI: I've applied these IAM tweaks with terraform both in `dev` and `alpha`.

**Comment by ntoll** [comment]
Squashed the commits.

File: scripts/version_buckets.py [hunk 8]
```diff
@@ -0,0 +1,243 @@
...
+    buckets = get_buckets(number)
+    for k, v in buckets.items():
+        name = k
+        bucket = v["bucket"]
+        versioning = v["versioning"] == "Enabled"
+        lifecycle = v["lifecycle"]
+        update_bucket(name, bucket, versioning, lifecycle, execute)
+
+
+def update_bucket(name, bucket, versioning, lifecycle, execute):
```
xoen, line 187: this `lifecycle` argument is not used in this function right?

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+        for i in range(3):
+            output += row[i].ljust(col_width[i])
+        if not VERBOSE:
+            click.echo(output)
+        logger.info(output)
+
+
+@main.command()
+@click.option("-n", "--number", required=False, type=int)
+@click.option("-x", "--execute", is_flag=True)
```
xoen, line 172: Is `--execute` like a `--dummy`/`--dry-run` opposite? I saw some code commented below, I wonder if that's intentional.

If we comment code because it has side effect it may mean we may want to default to `False` if it worries us.

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+    # Add versioning if not already set.
+    if not versioning:
+        msg = f"Enabling versioning for {name}."
+        logger.info(msg)
+        if not VERBOSE:
+            click.echo(msg)
+        if execute:
+            # TODO: Uncomment to actually make the changes.
+            # v = bucket.Versioning()
+            # v.enable()
```
xoen, line 204: This is the commented code I was referring to in other comment related to `--execute` flag.
If we default execute to `False` we may feel comfortable enough to have this uncommented (as it wouldn't run by default, unless of course the `--execute` flag is ignored).

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+            click.secho("OK", fg="green")
+        else:
+            click.secho("OK", fg="yellow")
+    # Set life cycle rule to send non-current versions of files to glacier
+    # storage after 30 days.
+    lifecycle_id = f"{name}_lifecycle_configuration"
```
xoen, line 210: Does the ID matter? I don't know much about S3 lifecycle configurations but I wonder if having the name of the bucket in it could cause some problems (e.g. name too long or something).

Additionally it may make sense to have a more descriptive name, e.g. `non_current_versions_to_glacier_after_30d` or something. 

Sorry I only thought about this now, to be honest if it works it doesn't matter, I'm just thinking aloud really ☺️

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+    msg = f"Setting lifecycle {lifecycle_id} for bucket {name}."
+    logger.info(msg)
+    if not VERBOSE:
+        click.echo("\n\n" + msg)
```
xoen, line 214: Shouldn't it be:

```python
if VERBOSE:
   click.echo(...)
```

Maybe I'm missing something. I see the `msg` is sent to the logger anyway so I'm assuming here that `VERBOSE` is to control additional output to the stdout?

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+                "NoncurrentVersionTransitions": [
+                    {"NoncurrentDays": 30, "StorageClass": "GLACIER",},
+                ],
+            },
+        ]
+    }
+    msg = json.dumps(life_cycle)
+    logger.info(msg)
+    if not VERBOSE:
+        click.echo(msg)
```
xoen, line 230: same `if` statement here so I must misunderstand how this `VERBOSE`/`--verbose` flag works sorry.

scripts/version_buckets.py
ntoll: Correct. Needs refactoring out.

scripts/version_buckets.py
ntoll: I just wanted to be explicit that the command should have side effects so I didn't make changes without being explicit about it.

scripts/version_buckets.py
ntoll: Yeah.... I won't uncomment this until ready. I just wanted to make sure anyone else running the command couldn't actually make any changes. Commenting out this (and the other) block of code ensures nothing accidental happens.

scripts/version_buckets.py
ntoll: It works. The name as it stands describes what it refers to and what it actually is (a lifecycle configuration). I'm calling a spade, a spade. If I were looking at resources in our AWS console and found this, I think I'd have a pretty good idea what it was, to what it pertained and what it represented. If I wanted to know what it did, I could just look at the definition. ;-)

scripts/version_buckets.py
ntoll: Nope. Verbose emits ALL logging out to stdout so `click.echo` is surplus to requirements here.

scripts/version_buckets.py
ntoll: Ditto :-)

ntoll:
Aldo... top review there matey. Thank you..! :-) Still some minor changes I want to make in light of the testing yesterday. I'll ping you when I've pushed the (minor) updates (which I was working on last thing yesterday).

scripts/version_buckets.py
ntoll: Actually, just looked again, it definitely needs to be in there. ;-) See line 196.

xoen: hang on, that's the `versioning` argument tho right? I still can't see the `lifecycle` argument being used int this function, maybe the caffeine hasn't triggered yet 😬(an unused argument is still shouldn't cause problems tho, better than a _missing_ argument 😋 )

scripts/version_buckets.py
xoen: We tend to call this `--dry-run` as it's the common name for flags like this and also to have it true by default so that it doesn't do things for real by default but I guess it doesn't matter ☺️

scripts/version_buckets.py
xoen: Oh I see, it only use echo when `VERBOSE` is false as when it's true it would output a lot of things anyway, now I'm with you 👌

scripts/version_buckets.py
davidread: No big deal, but I agree that the bucket name is redundant in the lifecycle ID. It's stored against the bucket, so no need to make it globally unique. Better that the ID front-loads what it does.

scripts/version_buckets.py
davidread: 4 lines for every log/output message. surely there's a better way?

What about log.info() outputs to the console, unless you specify -v when log.debug() also goes to the console? Or maybe there's some advantage of click.echo() that we'd miss?

scripts/version_buckets.py
```diff
@@ -0,0 +1,243 @@
...
+
+    See discussion here:
+
+    https://github.com/ministryofjustice/analytics-platform-ops/pull/320#discussion_r433765261
+    """
+    is_valid = name.startswith("alpha-app-")
+    if name.startswith("dev-"):
+        # Ignore this since it's a bucket associated with the dev instance.
+        is_valid = False
+    elif not is_valid:
```
davidread, line 67: I think there's unnecessary cognitive load in reading lines like this. What does 'valid' mean in this context? I get the impression you're working too hard to have a single return statement, either because there is a rule of thumb to achieve that, or because the log statement.
```
if name.startswith("alpha-app-"):
    # these are all the App Data Sources
    return True
if name.startswith("dev-"):
    # not interested in that environment
    return False
```
I think the 'single return statement' rule is not hard and fast:
https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement
and you could move the log statement to move to where you call this function.

xoen: I tend to agree with that.
I usually prefer to deal with the exceptional/trivial cases early on and leave the rest of the function for the main logic. It makes things more readable, e.g.

```python
if not name.startswith("alpha-"):
    # not interested in other environments
    return False

if name.startswith("alpha-app-"):
    # these are all the App Data Sources
    return True

# Rest of the logic which check `buckettype=datawarehouse` tag
# ...
```

I find this easier to read as I don't have to track down where/if the `is_valid` variable changed but I'm a bit slow so maybe it's just me 😋.

ntoll: In this case, the heuristics are not as simple as yes/no -- there's an "and" in the check in there somewhere (hence the need of the flag). Also, it seemed prudent to log the value of the flag so we had a record somewhere. The code does this at the end of this function before returning the result.

scripts/version_buckets.py
ntoll: Ack... I'll just make it `lifecycle_configuration`.

scripts/version_buckets.py
ntoll: Heh... depends on the tools I suppose. I've seen the `-x` (for `eXecute`) flag elsewhere. No big deal I agree... ;-)

scripts/version_buckets.py
ntoll: My bad. Fixed in latest push.

ntoll:
Merging since the updates are now underway as per the deploy plan.

PR #324, opened 2020-06-12
Add Cloudwatch alerts for Kubernetes master and bastion ELBs

andylightfoot:
1. Add a new terraform module that creates a 2 Cloudwatch alerts for an ELB
     - Alert on number of healthy hosts in an ELB target group
     - Alert on length of surge queue

2. Add an output from the existing Cloudwatch alerts module that contain the ARN of the SNS topic we send alerts to.

3. Apply these alerts to the Kubernetes master and Bastion ELBs in the alpha and dev environments


infra/terraform/platform/main.tf
```diff
@@ -205,3 +205,27 @@ module "buckets_archiver" {
     "component", "buckets-archiver",
   ), var.tags)}"
 }
+
+module "kubernetes_master_monitoring" {
+  source = "../modules/elb_cloudwatch_alerts"
+
+  name          = "${terraform.workspace}-kubernetes-master-alerts"
+  elb_name      = "${var.kubernetes_master_elb_name}"
+  alarm_actions = ["${module.softnas_monitoring.stack_notifications_arn}"]
```
xoen, line 214: Should the alarms for the masters go into the SoftNAS topic?

infra/terraform/platform/main.tf
```diff
@@ -205,3 +205,27 @@ module "buckets_archiver" {
     "component", "buckets-archiver",
   ), var.tags)}"
 }
+
+module "kubernetes_master_monitoring" {
+  source = "../modules/elb_cloudwatch_alerts"
+
+  name          = "${terraform.workspace}-kubernetes-master-alerts"
+  elb_name      = "${var.kubernetes_master_elb_name}"
```
xoen, line 213: I see in the variables that this value is `api-dev`/`api-alpha`, potentially you could get rid of the `kubernetes_master_elb_name ` variable in the tfvar file and use `elb_name      = "api-${terraform.workspace}"` here.

Similar below you could get rid of the `bastion_elb_name` variable and replace with `elb_name      = "bastion-${terraform.workspace}"`.

If you decide to not do this is also fine, I would personally slightly prefer to have 2 variables less in this case but it's not the end of the world anyway ☺️ .

andylightfoot: Yes good idea. I'll change them to use that

infra/terraform/platform/main.tf
andylightfoot: Yeah I think I'll probably eventually take the SNS notification stuff out of the SoftNAS module as it probably doesn't make sense for us to set up and new SNS topic for each set of alerts.

andylightfoot: I've create a new SNS topic for AP infra alerts and pointed the new alerts at this topic rather then reusing the softnas topic

xoen:
Nice one Andy ☺️

PR #325, opened 2020-06-15
Add Cloudwatch alerts for CPU usage of Kubernetes master instances

andylightfoot:
We want to send an alert of the CPU usage on the Kubernetes master nodes is high for a sustained period.

As the Kubernetes master nodes are in AutoScaling groups we need to put the monitoring on the Auto scaling groups rather than the individual EC2 instances as the instance IDs of these could change.

- Add a module to create an alert rule for high CPU usage in an Autoscaling group
- Use this module to create alerts for each of the Auto scaling groups used by the Kubernetes master nodes.

### DIFF
--- a/infra/terraform/modules/asg_cloudwatch_alerts/README.md
+++ b/infra/terraform/modules/asg_cloudwatch_alerts/README.md
@@ -28,6 +28,6 @@ Parameters
 | Name                                 | Type     | Description                               |
 | ------------------------------------ | -------- | ----------------------------------------- |
 | `name`                (**Required**) | `string` | Name of the resources |
-| `asg_name`            (**Required**) | `string` | Names of the ELB to monitor |
+| `asg_names`            (**Required**) | `string` | Names of the ELB to monitor |
 | `alarm_actions`       (**Required**) | `list`   | List of ARNS of resources for alerts to be sent to |
 | `tags`                (**Required**) | `map`    | Tags to attach to resources |
--- a/infra/terraform/modules/asg_cloudwatch_alerts/README.md
+++ b/infra/terraform/modules/asg_cloudwatch_alerts/README.md
@@ -1,0 +1,33 @@
+# CloudWatch Alerts
+
+##### Terraform module to enable CloudWatch alerts for an Auto Scaling Group
+
+Sends alerts when CPU usage goes above ${cpu_threshold} percent.
+
+Usage
+-----
+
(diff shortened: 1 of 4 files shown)


infra/terraform/modules/asg_cloudwatch_alerts/README.md
```diff
@@ -0,0 +1,33 @@
...
+  ), var.tags)}"
+}
+```
+
+Parameters
+-----------
+| Name                                 | Type     | Description                               |
+| ------------------------------------ | -------- | ----------------------------------------- |
+| `name`                (**Required**) | `string` | Name of the resources |
+| `asg_name`            (**Required**) | `string` | Names of the ELB to monitor |
```
xoen, line 31: `asg_names` ☺️ (missing the `s`) - sorry for the annoying typo comment 😬

infra/terraform/platform/main.tf
```diff
@@ -226,6 +226,18 @@ module "kubernetes_master_monitoring" {
   ), var.tags)}"
 }
 
+module "kubernetes_master_asg_monitoring" {
+  source = "../modules/asg_cloudwatch_alerts"
+
+  name          = "${terraform.workspace}-kubernetes-master-asg-alerts"
+  asg_names     = ["master-eu-west-1a.masters.${terraform.workspace}.mojanalytics.xyz", "master-eu-west-1b.masters.${terraform.workspace}.mojanalytics.xyz", "master-eu-west-1c.masters.${terraform.workspace}.mojanalytics.xyz"]
```
xoen, line 233: This is coming from Kops, that's why we don't have this list in Terraform, right? 👌

infra/terraform/modules/asg_cloudwatch_alerts/README.md
andylightfoot: Ah yes good spot!

andylightfoot: All fixed thanks!

infra/terraform/platform/main.tf
andylightfoot: Yeah that's right, we could maybe read from a data source but probably isn't worth it.

xoen: Agreed ☺️ 👌 - not ideal but at least it's simple ☺️

PR #327, opened 2020-06-16
Update Softnas cpu alerts so that only send alerts when we need to investigate

andylightfoot:
- Change CPU alarms to be based on Average CPU rather than maximum as we want to alert if sustained high CPU not small spikes
- Add variable for CPU alerts monitoring period and data point to trigger alarms so defaults can be over-ridden when required
- Add variable for CPU low usage threshold so that we can have a different value in dev and alpha
- Change expression used to define which SoftNAS instance is monitored so that we don't add alerts for the stopped SoftNAS instance (this is a bit of a hack but there isn't an easy way of renaming the active SoftNAS instance in alpha. This logic works for both dev and alpha.


infra/terraform/platform/main.tf
```diff
@@ -57,10 +57,10 @@ module "softnas_monitoring" {
   source = "../modules/cloudwatch_alerts"
 
   name               = "${terraform.workspace}-softnas-alerts"
-  ec2_instance_ids   = "${module.user_nfs_softnas.ec2_instance_ids}"
-  ec2_instance_names = "${module.user_nfs_softnas.ec2_instance_names}"
+  ec2_instance_ids   = ["${element(module.user_nfs_softnas.ec2_instance_ids, 1)}"]
```
xoen, line 60: So this is only passing in the second EC2 instance. 
**Maybe** worth adding a comment on top of that line to explain we're not interested in the first instance as it's not in use.

This is also fine in `dev`? I think we have one EC2 instances for SoftNAS in `dev`, isn't this going to go out of index? Maybe not, (I didn't test this by running `terraform`)

andylightfoot: Yes it work in dev as it will just match the first in the instance (lists are a bit weird in terraform). It is a bit hacky but the least bad solution I could come up with for now. I'll add a note to explain

xoen: Oh OK, I wasn't sure because as you said, Terraform loops are...special 😬 and given indexes start from `0` I thought it would cause some problems when you do something like `element(..., 1)` but apparently not 🤷‍♂️ (weird!)

andylightfoot: Yeah so what happens if it reaches the end of the list it matches from the start again....a bit weird but handy in this case!

xoen: ahahah

PR #333, opened 2020-07-02
Update kops config to match the current kops state

andylightfoot:
The kops configuration in this repo does not reflect the current config in the kops state bucket or the clusters.

It looks like the kops state has been updated directly (rather than updating from the terraform output) for a least a year.

This PR aims to retrofit some of these changes.

- Update Kubernetes version
- OIDC client ID
- Node pool instance types
- Node pool min size


infra/terraform/modules/kops_spec/main.tf
```diff
@@ -108,7 +108,7 @@ data "template_file" "bastions_instancegroup" {
     machine_type              = "${var.bastions_machine_type}"
     root_volume_size          = "${var.bastions_root_volume_size}"
     max_size                  = "${var.bastions_instancegroup_max_size}"
-    min_size                  = "${var.bastions_instancegroup_max_size}"
```
damacus, line 111: 🤨 what now

infra/terraform/platform-base/variables.tf
```diff
@@ -25,6 +25,10 @@ variable "oidc_provider_url" {}
 
 variable "oidc_client_id" {}
```
damacus, line 26: Is this one still used?

infra/terraform/modules/kops_spec/main.tf
andylightfoot: The min size of the IG was set to equal the value of the max_size variable, we don't want that we want it to be the min_size variable

infra/terraform/platform-base/variables.tf
andylightfoot: Yep it is used by the federated_identity module (which I think sets up the saml provider stuff in AWS) and is the correct value.

infra/terraform/modules/kops_spec/main.tf
andylightfoot: Min size was set to be the value of the max_size variable, instead of the min_size variable.

infra/terraform/platform-base/variables.tf
andylightfoot: Yep it is used by the federated_identity module (is using the correct value)

xoen:
@damacus it is a bit of an "annoyance" that we can't see these in GitHub but it's probably better than having them in clear in a public PR 😬 

I personally usually check out the branch under review and have a look at the (decrypted) `*.tfvars` files/diff locally - not as convenient as looking at this on GH but better than nothing ☺️

infra/terraform/platform-base/main.tf
```diff
@@ -62,7 +62,7 @@ module "kops_spec" {
   cluster_dns_name = "${module.cluster_dns.dns_zone_domain}"
   cluster_dns_zone = "${module.cluster_dns.dns_zone_id}"
 
-  oidc_client_id  = "${var.oidc_client_id}"
+  oidc_client_id  = "${var.kubernetes_oidc_client_id}"
```
xoen, line 65: I've checked the `tfvars` file and the Client ID looks correct 👍 

I'm not familiar with the other `oidc_client_id` value so not sure whether is used or not or how.

...oh I just saw [your comment](https://github.com/ministryofjustice/analytics-platform-ops/pull/333/files#r449076123) 👌

infra/terraform/modules/kops_spec/main.tf
damacus: Yeah I was just wondering why the heck it was set as that!

PR #335, opened 2020-07-07
Increase size of root volume in alpha cluster nodes to 150GB

andylightfoot:
Currently we allocate 100GB to the root volume of the kubernetes nodes in our alpha cluster. We have seen that pods have been evicted from a node due to the disk filling up (and several warning alerts about disk space from other nodes). The vast majority of the disk space is taken up by Docker images that have been pulled to the node. We should increase the size of the volume of our nodes to ensure that we don't run out of disk space.

This change increases the size of the root volume of nodes to 150GB.


damacus:
Couple of things: 
- Any reason why we can't just double to 200gb? 
- What kind of Docker image pruning re we doing on the nodes?

andylightfoot:
> Couple of things:
> 
> * Any reason why we can't just double to 200gb?
> * What kind of Docker image pruning re we doing on the nodes?

We can double to 200GB but I'm not convinced that is necessary and I like saving money for the taxpayers. We can always increase later on if needed and it might put pressure on us to descrease the size of some of the images (and reduce the number of different version of R Studio people are using)

andylightfoot:
> Couple of things:
> 
> * Any reason why we can't just double to 200gb?
> * What kind of Docker image pruning re we doing on the nodes?

Re image pruning, I think this is just using the standard K8s GC settings

PR #339, opened 2020-07-07
Upgraded nginx-ingress-controller to latest beta and replaced custom …

damacus:
…nginx conf template with force-ssl-redirect annotation


andylightfoot:
Could you put some more details on the changes?

damacus:
I just found this PR lying around and opened it up for discussion.

davidread:
So old, let's let it go

PR #340, opened 2020-07-08
Add kube-reserved and system-reserved to kops

andylightfoot:
We have seen Kubernetes nodes being starved of CPU and memory for system processes due to all the resources being allocated to kubernetes workloads.

This change will allow us to reserve some resources for kubernetes and system resources (such as docker). This will result in a reduction in the amount of Allocatable resources on a node (so we will need to run more nodes as a result, but they should be more stable).

https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

**Things that I'm not sure about**
- I'm not sure that we need to allocate both kube-reserved and system-reserved resources (as aren't creating cgroups for each of these) but I think it might be useful to specify both.

- Our current alpha nodes have a total of 8 CPUs, 64Gb of memory and 150Gb of storage.
I have reserved a total of 1.5 CPUs, 4Gb of memory and 7Gb of storage for kube/system resources on our alpha nodes. I think this should be more than sufficient but I am open to feedback on this.


damacus:
## Memory

- This seems like a really good start. We can always increase/decrease. We have enough spare nodes for this to work as is

## Kube and system reserved

I've got nothing to say this is OTT right now. Same as before, lets tune going forward with these settings 👍

PR #347, opened 2020-08-25
We should get GPG keys from GitHub.

davidread:
We're transitioning away from storing gpg keys in https://github.com/ministryofjustice/analytical-platform-public-keys

Related PR: https://github.com/ministryofjustice/analytical-platform-public-keys/pull/9


ntoll (approved):
LGTM :+1:

PR #348, opened 2020-09-07
Remove old S3 buckets

damacus:
Remove crest, scratch and source buckets
Remove their associates iam groups and group policies
Remove the associated module

Add the versions file to tfswitch choses the correct version of Terraform to run (0.11)


damacus:
Confirmed with @RobinL that these buckets are no longer required.

infra/terraform/modules/data_buckets/s3.tf
```diff
@@ -1,101 +0,0 @@
-resource "aws_s3_bucket" "source" {
```
xoen, line 1: FYI: I don't know much about this `"source"` S3 bucket.

xoen (approved):
Nice spring cleaning @damacus 🌟 

AFAICT and remember these should be safe to remove especially if you also asked Robin.

I don't know much about the "source" one and the CREST one probably is not used as they possibly moved to the new way of working with data (raw/curated/etc managed by DE team) - and you checked with Robin so 👍

PR #349, opened 2020-09-07
Stop passing the workspace name to submodules

damacus:
- Call terraform.workspace directly in modules
- Remove the now defunct lambda_functions module
- Split the provider configuration out into it's own file
- Rename inputs.tf to variables.tf where ever it doesn't match


damacus:
This is a 0-diff change with terraform plan

tomskelley-gjs (approved):
LGTM 👍

PR #351, opened 2020-09-07
Remove the log_pruning module

damacus:
We no longer use this resource as it hasn't ever worked


tomskelley-gjs (commented):
LGTM - this never worked anyway.

damacus:
Also I already deleted it 🤷

tomskelley-gjs (approved):

PR #353, opened 2020-09-07
Remove variables related to atlantis

damacus:
Atlantis was removed a while ago. We no longer use these variables. This
will be an error in future versions of Terraform


tomskelley-gjs (approved):
LGTM

PR #354, opened 2020-09-07
Move Terraform modules to the global folder

damacus:
Moves Terraform modules used by the global Terraform, to a subfolder of
global Terraform.

This makes it easier to segregate global and environment resources.


tomskelley-gjs (approved):
LGTM

PR #355, opened 2020-09-07
Remove the EBS snapshot & pruning modules

damacus:
Neither of these modules work correctly.

Pruning has never worked
Snapshot creation constantly errors on "some" volumes as we are creating
them too fast. We will replace these modules with a different approach.


andylightfoot:
I've had a quick look and it looks like the EBS Snapshot Lambda is successfully creating some snapshots of the etcd volumes currently.  So I think it would be better to implement the replacement first before deleting what we have already.

damacus:
Converting to a draft whilst we come up with a plan for a replacement 👍

andylightfoot:
> Converting to a draft whilst we come up with a plan for a replacement 👍

Sounds good

damacus:
Looks like this module of @xoen's is actually what we want long term. 

https://github.com/ministryofjustice/analytics-platform-ops/pull/301

PR #356, opened 2020-09-08
Move IAM policies from inline to iam_policy_documents

damacus:
Move IAM policies from inline to iam_policy_documents


xoen (approved):
LTGM 👍 

Nice one Dan ☺️ 🌟

PR #359, opened 2020-09-08
Add oidc provider thumbprints for new Auth0 CA cert

andylightfoot:
On 15th Sep 2020 Auth0 are going to be updating the certs for our auth0 tenant. As a result we need to update the AWS IAM Identity provider with thumbprint for new Auth0 CA cert.

This change adds the certificate thumbprint for new Auth0 CA cert for auth0 alpha domain OIDC provider


damacus:
What's the best way of checking if this thumbprint is correct?

andylightfoot:
> What's the best way of checking if this thumbprint is correct?

I don't think we can until the cert changes. We just have to trust that the thumbprint that Auth0 sent us is correct.

damacus (approved):
Brill. If this is just a copy 'n pasta from an email. I'm happy with it.

PR #360, opened 2020-09-09
Remove env parameter as has been removed from these modules

andylightfoot:
The env parameter had been removed from the modules but not the calling code.


tomskelley-gjs (approved):
LGTM

damacus:
Ace. I thought I got them all!

PR #363, opened 2020-09-09
Add vpcflowlogs lambda to the assets folder

damacus:


tomskelley-gjs (approved):
I can't review the python as I'm not a python person, but seems sane from a terraform perspective...

damacus:
Yeah it's a straight copy from the other folder. Nothing's been changed.

PR #365, opened 2020-09-10
Add tags to resources & tidy up the lambda_mgmt module

damacus:
## What

This PR;

- Adds the mandatory tags to all available resources
- Add descriptions to variables so `terraform-docs` can generate the table for the `lambda_mgmt` module correctly 
- Adds types to variables so variables don't just let anything through
- Uses Terraform 0.12 handling of arrays to pull back an array from `aws_route53_zone.global.name_servers` rather than specifying all the name servers in the array manually. This makes moving between name servers/ regions easier.

### DIFF
--- a/infra/terraform/global/main.tf
+++ b/infra/terraform/global/main.tf
@@ -34,17 +34,27 @@ module "aws_account_logging" {
   vpcflowlogs_s3_bucket_name = var.vpcflowlogs_s3_bucket_name
 
   vpc_id = var.vpc_id
+  tags   = local.tags
 }
 
 module "mojanalytics_concourse_iam_list_roles_user" {
   source      = "./modules/iam_list_roles"
   org_name    = "mojanalytics"
   system_name = "concourse"
+  tags        = local.tags
 }
 
 module "ses_domain" {
-  source = "./modules/ses_domain"
-  domain = var.platform_root_domain
-
+  source              = "./modules/ses_domain"
+  domain              = var.platform_root_domain
   aws_route53_zone_id = aws_route53_zone.platform_zone.zone_id
+}
+
+locals {
+  tags = {
+    business-unit = "Platforms"
+    application   = "analytical-platform"
+    owner         = "analytical-platform:analytics-platform-tech@digital.justice.gov.uk"
+    is-production = "true"
+  }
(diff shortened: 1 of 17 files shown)


infra/terraform/global/main.tf
```diff
@@ -34,17 +34,27 @@ module "aws_account_logging" {
...
+  domain              = var.platform_root_domain
   aws_route53_zone_id = aws_route53_zone.platform_zone.zone_id
 }
+
+locals {
+  tags = {
+    business-unit = "Platforms"
+    application   = "analytical-platform"
+    owner         = "analytical-platform:analytics-platform-tech@digital.justice.gov.uk"
+    is-production = "true"
```
xoen, line 58: Should `is-production` always be `"true"`?

Several resources also have the `env` tag (which currently can be `env="dev"` or `env="alpha"`).

You could have a look at the `platform` variables to have an idea of what I mean.

infra/terraform/global/modules/lambda_mgmt/CHANGELOG.md
```diff
@@ -1,9 +0,0 @@
-# Changelog
```
xoen, line 1: How is it that the Changelog got dumped? Wasn't it up-to-date? I still find changelogs very useful and quicker than checking the git commits but maybe I'm missing something specific of this module.

xoen (commented):
From what I can see and understand this looks good but I left some thoughts on the tags.

infra/terraform/global/modules/lambda_mgmt/CHANGELOG.md
damacus: The CHANGELOG for the module was pretty much empty and wrong. 

- We don't version the module
- We can't version the module

infra/terraform/global/main.tf
damacus: Yeah, good point. This really does need to be a switch. I should be able to pick this up when it's applied.

damacus: On second thoughts. I've just reread the code. There is only one workspace `default`, and all of the resources that are being tagged here are actually production in some way.

PR #368, opened 2020-09-30
CP Redis cluster on AWS

xoen:
### What

This creates a Redis cluster (with "cluster mode" disabled) in AWS ElastiCache to be used by the Control Panel.

### Why?
Currently Redis runs in a pod in our k8s (kubernetes) cluster. CP Redis mostly contains session data and background jobs queue but nevertheless it's not good to run stateful application on k8s.

Furthermore, this k8s redis is [deployed using an helm dependency](https://github.com/ministryofjustice/analytics-platform-helm-charts/blob/414d5e7aae2db8ea9b10dfd803f9abc53f4d1a13/charts/cpanel/requirements.yaml#L7-L9), these are not updated and have caused issues in the past (not for the `cpanel` release itself IIRC but we've seen other cases where `helm upgrade` were failing because of some of the dependencies k8s resources confusing it).

In general we want to avoid maintaining DBs and things like that if there is a managed service we can use instead as it reduce the maintenance load on our team.

### Why "cluster mode" disabled?
The name "cluster mode" is a bit confusing so let's start with that.

See this document please: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis-RedisCluster.html

"cluster mode: disabled" doesn't necessarily mean only `1` redis node with no replication. It's a new way of splitting data across different nodes, e.g. "sharding" it so that each node doesn't contain the full data. This has good scaling/availability/performances implications on paper.

Unfortunately, it looks like both [`django-redis`](https://github.com/jazzband/django-redis/issues/208) and [Django Channels](https://github.com/django/channels/issues/485) **don't support "cluster mode: enabled", that's why we don't use it here** (cluster mode disabled = 1 shard).

That sounds bad but the reality is that this redis cluster will still:
- have 3 nodes, `1` primary and `2` replicas
- will have Auto-failover enabled, so if the primary fails for whatever reason, one of the replicas will become the new primary
- has redundancy so doesn't rely on a single node to be working
- nodes are in 3 AZs (Availability Zones) so in theory can survive even an AZ being down

We probably don't need this level of availability/redundancy but it's good to have if it can reduce the CP availability.

### Terraform documentation
Links to documentation on Terraform resources used:
- [`elasticache_replication_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group)
- [`aws_elasticache_subnet_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_subnet_group)
- [`aws_security_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group)

### Redis Instance Type

We've decided to (at least for now):
- use `cache.t3.micro` for `dev` (2 vCPU / 0.555 GB memory)
- use `cache.t3.medium` for `alpha` (2 vCPU / 3.09 GB memory)

See [AWS Redis pricing page](https://aws.amazon.com/elasticache/pricing/) for instance types specs, prices, etc...

### Other
I've also added some tagging to new and existing Terraform resources here as that's a best practice and it helps keep track of what resource belong to what which is always welcome when removing things down the line.

### Ticket
https://trello.com/c/YG9anK8A


infra/terraform/modules/control_panel_api/redis.tf
```diff
@@ -0,0 +1,40 @@
...
+
+resource "aws_security_group" "control_panel_redis" {
+  name   = "${terraform.workspace}-control-panel-redis"
+  vpc_id = "${var.vpc_id}"
+
+  ingress {
+    from_port       = "${var.redis_port}"
+    to_port         = "${var.redis_port}"
+    protocol        = "tcp"
+    security_groups = ["${var.ingress_security_group_ids}"]
```
xoen, line 31: This will be `extra_node_sg_id`, from module client code:

```HCL
ingress_security_group_ids = ["${data.terraform_remote_state.platform_base.extra_node_sg_id}"]
```

xoen:
FYI: I got the CP in `dev` to connect to this Redis cluster in AWS EC thanks to [this tweak to the connection settings](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/843).

xoen:
Any objections if this get applied in `alpha` as well so that we have the AWS EC Redis cluster ready for the move once other PRs are approved? As said, in `dev` this worked once I told CP to connect using SSL/TLS so in theory should work.

infra/terraform/modules/control_panel_api/redis.tf
```diff
@@ -0,0 +1,40 @@
+resource "aws_elasticache_replication_group" "control_panel_redis" {
+  replication_group_id          = "${terraform.workspace}-control-panel-redis"
+  replication_group_description = "${terraform.workspace} CP Redis cluster"
+
+  engine         = "redis"
+  engine_version = "5.0.6"
```
andylightfoot, line 6: might be worth making this a variable so we can more easily upgrade at later time?

andylightfoot (approved):
Maybe worth making the redis version use a variable? Otherwise LGTM.

infra/terraform/modules/control_panel_api/redis.tf
xoen: Cool, I'll move to a variable and default to `5.0.6`. In theory AWS will auto-upgrade the minor versions anyway which I guess will confuse Terraform a little but that's always the case with DB versions I think...

Anyway, thanks for the review @andylightfoot ❤️  - I'll make this change and merge/apply in `alpha` so at least the Redis cluster will be ready and I can get the endpoint and put in the config 👌

xoen: @andylightfoot FYI: [added `redis_engine_version` variable](https://github.com/ministryofjustice/analytics-platform-ops/pull/368/files#diff-0c4b2b1e7fb9bdf2271e8ea91829d9faR47-R49) and applying to `alpha` now...It's a bit slow at spawning a new Redis cluster but I don't imagine it failing but I'll merge after that.

PR #372, opened 2020-11-10
Add script to update s3 versioning related permissions for existing users.

ntoll:
## What

This script gives existing users of the control panel the permissions they need to work with versioned S3 buckets.

## How to review

1. Test on a `dev` user and ensure the error message for the related s3 bucket no longer appears in the web based admin page.

Trello reference: https://trello.com/c/HqDQMr6s/843-fix-s3-user-permissions

### DIFF
--- a/scripts/upgrade_s3_permissions.py
+++ b/scripts/upgrade_s3_permissions.py
@@ -9,6 +9,8 @@ Requires boto3 and click packages to be installed.
 
 What does this script do?
 
+This script ensures that every user (or a named user) has the expected
+permissions for working with versioned S3 buckets.
 """
 import os
 import click
@@ -85,8 +87,9 @@ def update_user(username, execute=False):
     response = iam_client.get_role_policy(RoleName=username,
                                           PolicyName='s3-access')
     policy_document = response["PolicyDocument"]
+    # TODO: Check the permissions are not already correct.
     if execute:
-        # Patch the thing
+        # TODO: Patch the thing
         response = client.put_role_policy(RoleName=username,
                                           PolicyName='s3-access',
                                           PolicyDocument=new_policy_document)
--- a/scripts/upgrade_s3_permissions.py
+++ b/scripts/upgrade_s3_permissions.py
@@ -17,6 +17,7 @@ import click
 import logging
 import sys
 import boto3
+from botocore.exceptions import ClientError
 import json
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


scripts/upgrade_s3_permissions.py
```diff
@@ -0,0 +1,163 @@
...
+    "s3:GetBucketPolicy",
+    "s3:GetBucketAcl",
+    "s3:GetBucketCORS",
+    "s3:GetBucketVersioning",
+    "s3:GetBucketLocation",
+    "s3:ListBucketVersions"
+]
+
+LISTUSERBUCKETS_SECTION = {
+    "Sid": "ListUserBuckets",
```
xoen, line 52: Again, the `Sid` here seemed to have changed from `console` to `ListUserBuckets`. Maybe it's more accurate, but it's a risk to introduce this change.

scripts/upgrade_s3_permissions.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
```
xoen, line 1: This code is quite far from the corresponding Control Panel code dealing with AWS (which I believe is somehow duplication, but I don't mean this is as a bad thing): https://github.com/ministryofjustice/analytics-platform-control-panel/pull/854/files

I wonder if it could be closer to that code and be a django command instead or even migration (probably a bad idea), potentially even reusing some of the code there and/or constants.

You'd have the additional benefit of being part of the same codebase and potentially be easier to discover/modify when AWS change these again (hopefully never and/or in years 😬 )

Just an idea, I know it's not the important part here ☺️

scripts/upgrade_s3_permissions.py
```diff
@@ -0,0 +1,163 @@
...
+    for statement in policy_document["PolicyDocument"]["Statement"]:
+        # Update the "Action" list depending on the statement ID (Sid).
+        sid = statement["Sid"]
+        if sid == "console":
+            statement = LISTUSERBUCKETS_SECTION 
+        elif sid == "readwrite":
+            statement["Action"] = READWRITE_ACTION
+        elif sid == "list":
+            statement["Action"] = LIST_ACTION
+        new_statements.append(statement)
```
xoen, line 146: If I'm reading this correctly...

The statements with Sid `console`, `readwrite` and `list` are updated/manipulated the rest is left intact, is that correct?

If that's the case, I see that `readwrite` statement is updated. So far so good. But the `readonly` one is not here, hence it would be left as before right?

I'm asking because [in twin PR's diff](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/854/files) I see that the `WRITE_ACTIONS` "constant" is not updated, but the `READ_ACTIONS` one is, to add:

```
    "s3:GetObjectVersionAcl",
    "s3:GetObjectVersionTagging"
```

Don't this script need to add them to the `readonly` statement too? 

PS: I may miss something fundamental here so apologies in advance if I'm telling something stupid and/or I'm missing something here.

xoen (commented):
I've mainly focused on the IAM bits and not much on the `click`/high level logic as I'm assuming it's working.

I've added a couple of comments/questions.

As said in [other PR](https://github.com/ministryofjustice/analytics-platform-control-panel/pull/854/files), I'm wary of renaming any of the IAM policy statemens Sids as they may be used programmatically (in fact they're used by this very script itself) so change them may break things in surprising and/or potentially dangerous ways.

The other important point/question I raised/asked is regarding the `readonly` statement which AFAICT is not updated but from the look of the other PR seems to need changing. So it may be worth double checking that before YOLO-run this in `alpha` ☺️

scripts/upgrade_s3_permissions.py
damacus: IMHO, I think this is quite a small risk here. One positive from this change, is we can see if users have been upgraded

scripts/upgrade_s3_permissions.py
ntoll: Ack, but I was told (for example when I added versioning to S3 buckets) that such migration scripts relating to infrastructure related settings should be in the `ops` repository. It feels to me like this is the right place, and we should not entangle ourselves with Django since this (ultimately) has nothing really to do with AWS related configuration.

scripts/upgrade_s3_permissions.py
xoen: But why risk it for a biscuit? The old `console` Sid was quite meaningful as it's clear this statement is there to allow AWS Console access. 

Plus, as I said in another comment, it seems like this `Sid` is actually used programmatically, e.g. at least in here on line `140`, so change it makes this script less idempotent  - reason being once this `Sid` is updated if you want to re-run the script to update these permissions (e.g. because we realise we need another permission) it would not update this statement anymore.

I strongly advice against changing this `Sid` as at best is a cosmetic change at worse can cause real problems so the risk/reward doesn't look great - IMHO.

damacus: Meaningful, but wrong. For what it's worth the *only* place the Sid is used is in this script.

scripts/upgrade_s3_permissions.py
damacus: We've now amended the `readonly` statement 👍

PR #373, opened 2020-11-12
Script to update webapps to use fluent-bit image in ECR

xoen:
### Why
This is to reduce the number of pull requests we're making to DockerHub
which is causing problems because of the newly introduced rate limites.

I also ran this through https://www.shellcheck.net/

Thanks to @damacus for the help with the Bash function.

### Ticket
https://trello.com/c/Izxt87HF

### DIFF
--- a/scripts/update-webapps-fluent-bit-image-to-ECR.sh
+++ b/scripts/update-webapps-fluent-bit-image-to-ECR.sh
@@ -1,0 +1,36 @@
+#! /bin/bash
+
+
+NAMESPACE=apps-prod
+CONTAINER=fluent-bit
+OLD_IMAGE=fluent/fluent-bit:1.1.1
+NEW_IMAGE=593291632749.dkr.ecr.eu-west-1.amazonaws.com/fluent-bit:1.1.1
+
+
+function uses_old_image() {
+    local deployment=${1:?}
+
+    kubectl -n ${NAMESPACE} get deployment "${deployment}" -o yaml | grep -q "image: ${OLD_IMAGE}"
+}
+
+## NOTE: You can also replace the following with a newline-separated list of
+##       `Deployments` to update, e.g.
+#
+# DEPLOYMENTS="
+# deployment-to-update-1
+# deployment-to-update-2
+# deployment-to-update-3
+# "
+DEPLOYMENTS=$(kubectl -n ${NAMESPACE} get deployment | grep \\-webapp | cut -d" " -f1)
+
+for deployment in $DEPLOYMENTS; do
+    echo
+    echo "Processing '${deployment}'..."
+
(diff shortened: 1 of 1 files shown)


scripts/update-webapps-fluent-bit-image-to-ECR.sh
```diff
@@ -0,0 +1,30 @@
...
+
+
+DEPLOYMENTS=$(kubectl -n ${NAMESPACE} get deployment | grep \\-webapp | cut -d" " -f1)
+
+for deployment in $DEPLOYMENTS; do
+    echo
+    echo "Processing '${deployment}'..."
+
+    if uses_old_image "${deployment}"; then
+        # kubectl -n "${NAMESPACE}" set image "deployment/${deployment}" ${CONTAINER}=${NEW_IMAGE}
```
andylightfoot, line 25: Is this meant to be commented out?

xoen: Nope.

andylightfoot: ok

andylightfoot (approved):

scripts/update-webapps-fluent-bit-image-to-ECR.sh
```diff
@@ -0,0 +1,30 @@
...
+
+    kubectl -n ${NAMESPACE} get deployment "${deployment}" -o yaml | grep -q "image: ${OLD_IMAGE}"
+}
+
+
+DEPLOYMENTS=$(kubectl -n ${NAMESPACE} get deployment | grep \\-webapp | cut -d" " -f1)
+
+for deployment in $DEPLOYMENTS; do
+    echo
+    echo "Processing '${deployment}'..."
```
damacus, line 22:
```suggestion
    echo "\nProcessing '${deployment}'..."
```

This is slightly more "Bash Like"  :) 

But LGTM!

damacus (approved):

scripts/update-webapps-fluent-bit-image-to-ECR.sh
xoen: Unfortunately don't think `echo` interprets those `\n` as you'd normally expect (I knew that but I re-tried and it didn't work, it just prints `\n` 😬 ).

xoen:
FYI: I've run this in `alpha`. I've added [a comment](https://github.com/ministryofjustice/analytics-platform-ops/pull/373/files#diff-9cd8fa57b85af69e1e428d356094e6f18ff38fa1d0bde8e3464c8faff0e12620R16) to the script to show how you can use it to update only a whitelist of deployments, e.g. to run things in batches like I did. Merging this.

PR #376, opened 2021-04-16
ANPL-390 Create new bastion servers

andylightfoot:
Create new Bastion servers in the dev and alpha VPC using the https://github.com/Guimove/terraform-aws-bastion terraform module.

It provisions an NLB and ASG in the DMZ subnets.

It has the existing bastion key assigned to it and the same security groups.

It uses the latest Ubuntu LTS AMI from Canonical.


tomskelley-gjs (approved):
LGTM

PR #377, opened 2021-04-19
use environment specific resource names for the bastions

andylightfoot:
Define environment specific variables for bastion_launch_template_name and bastion_iam_policy_name so that we avoid name conflicts.


neillturner (approved):
just alignment

tomskelley-gjs (approved):
LGTM

PR #381, opened 2021-04-27
Increase max nodes to 150

andylightfoot:
We are sometimes reaching the max number of nodes allowed for our alpha kubernetes cluster.

Increase the max number of nodes in the alpha cluster autoscaling group from 100 to 150.


tomskelley-gjs (commented):
LGTM

PR #390, opened 2021-11-16
Remove the arch docs in favour of the tech docs.

rossjones:
## What

The architecture documentation is now held in the [AP Tech Docs](https://ministryofjustice.github.io/ap-tech-docs) and so this
removes some duplication.


ntoll (approved):
LGTM :shipit:

